Issue #26, opened 2016-07-18
Reason for not using the node image?


stirringhalo:
Is there a particular reason you're not using the node docker image as a "from"?

klokan:
Debian is required esp. for installing xvfb for software GL rendering (typically used in Docker now).

We use debian as base of all our company docker containers, so it is more efficient (our server nodes downloads less data from docker hub).

What for advantages do you see in using official `node` as the base image?

If there are real benefits, dicsussed here and confirmed - would you be able to prepare a pull request?

stirringhalo:
Right, good point! I'm just leary of "curl | bash" as a regular method of installs. There's many ways it can break (incomplete transfers leaving half broken scripts running), DNS hijacking, etc. See stuff like: https://jordaneldredge.com/blog/one-way-curl-pipe-sh-install-scripts-can-be-dangerous/

The real repository is at:

deb https://deb.nodesource.com/${NODEREPO} ${DISTRO} main
deb-src https://deb.nodesource.com/${NODEREPO} ${DISTRO} main

where NODEREPO=node_4.x and I believe DISTRO=stretch or jessie

Just put those two lines into sources.list.d and be done with it? Sure nodesource could be equally malicious, but at least we're only running the scripts that need to be.

klokan:
+1. Pull request passing the tests?

stirringhalo:
Sure thing, I'll do that tonight!

klokan:
Great! Thanks.

BTW we are also working on bringing the code of "light" version into this repo - see #27 
The code may still change a bit - at least until announcement of version 1.0 on FOSS4G 2016.

petrsloup:
I believe this can be closed now (solved by #26).

PR #28, opened 2016-07-18
Avoid curl | bash

stirringhalo:
This avoids the arguably risky "curl | bash" with simply adding the repositories directly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,14 +3,20 @@ MAINTAINER Petr Sloup <petr.sloup@klokantech.com>
 
 RUN apt-get -qq update \
 && DEBIAN_FRONTEND=noninteractive apt-get -y install \
+    apt-transport-https \
     curl \
     unzip \
     build-essential \
     python \
     libcairo2-dev \
+    nodejs \
     xvfb \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,14 +3,20 @@ MAINTAINER Petr Sloup <petr.sloup@klokantech.com>
 
 RUN apt-get -qq update \
 && DEBIAN_FRONTEND=noninteractive apt-get -y install \
+    apt-transport-https \
     curl \
     unzip \
     build-essential \
     python \
     libcairo2-dev \
+    nodejs \
```
klokan: Really nodejs already here?

klokan:
@petrsloup please review / accept or comment.

Dockerfile
stirringhalo: You're right, that accidentally got left in after testing which version of nodejs gets installed into stretch. 7f06f09 removes it.

petrsloup:
Yes, this makes sense, works correctly and closes #26. Merged.

petrsloup:
Thanks for the contribution @stirringhalo !

PR #36, opened 2016-08-05
new feature: command line option for bind address

efi:
I ported this PR direktly in github, without checking the coude out / testing, but it should be OK, I think.

I created this github "organization" to be able to fork the original tileserver, since I got:

"You already have a fork of this repository: efi/tileserver-gl-light"

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -10,7 +10,7 @@ var opts = require('nomnom')
   })
   .option('bind', {
     abbr: 'b',
-    default: "::",
+    default: undefined,
     help: 'Bind address'
   })
   .option('port', {
--- a/src/main.js
+++ b/src/main.js
@@ -8,6 +8,11 @@ var opts = require('nomnom')
     default: 'config.json',
     help: 'Configuration file'
   })
+  .option('bind', {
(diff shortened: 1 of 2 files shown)


klokan:
Thanks! We will review, test and merge.

src/main.js
```diff
@@ -8,6 +8,11 @@ var opts = require('nomnom')
     default: 'config.json',
     help: 'Configuration file'
   })
+  .option('bind', {
+    abbr: 'b',
+    default: "::",
```
petrsloup: Please use single-quotes (`'`) to match the code style of the rest of the application.
Or even better, it might be a good idea to use `default: undefined` and let the `app.listen` do the logic:
```
If the hostname is omitted, the server will accept connections on any IPv6 address (::) when IPv6 is available, or any IPv4 address (0.0.0.0) otherwise.
```

petrsloup:
@efi I tested this and added one minor comment, otherwise LGTM.

efi:
Thanks, that is indeed the best solution.

petrsloup:
@efi Great, thank you for the contribution!

PR #69, opened 2016-11-02
remove some data from tileJSON

huangyingjie:
useless and sensitive properties.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -80,8 +80,13 @@ module.exports = function(options, repo, params, id) {
 
   app.get('/' + id + '.json', function(req, res, next) {
     var info = clone(tileJSON);
-    info.tiles = utils.getTileUrls(req, info.tiles,
-                                   'data/' + id, info.format);
+    info.tiles = utils.getTileUrls(req, info.tiles, 'data/' + id, info.format);
+    // delete usless info
+    delete info.Layer;
+    delete info.id;
+    delete info.filesize;
+    delete info.basename;
+    delete info.mtime;
     return res.send(info);
   });
 
--- a/test/metadata.js
+++ b/test/metadata.js
@@ -18,7 +18,7 @@ var testTileJSONArray = function(url) {
   });
 };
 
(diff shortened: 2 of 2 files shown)


klokan:
Thanks for the contribution.

Do you know that without `Layer` property the default viewers would not run at all?

If there are sensitive / invisible vector tiles to expose via TileJSON from your server - I suggest to implement "invisible" tag in the config.json for such layer - which will not show the layer in the list of data layers - and another "sensitive" tag which will hide the Layer list or other sensitive properties from TileJSON.

Direct removing of all the data for all layers makes no sense - it makes more sense to implement it as a config.json option. Please adjust this pull request.

Also ensure that the automated Travis tests are passing.

huangyingjie:
What's "default viewers"?
I test the tileJson at localhost, it's all right. And I read the source code of mapbox-gl-js, I have found no place using "Layer".
And see [this:](https://api.mapbox.com/v4/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v7.json?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpbG10dnA3NzY3OTZ0dmtwejN2ZnUycjYifQ.1W5oTOnWXQ9R1w8u3Oo1yA&secure)
It also doesn't contain Layer、basename..

petrsloup:
@huangyingjie Thanks for the PR. I see your point, but I did the proposed changes slightly differently (see https://github.com/klokantech/tileserver-gl/commit/2d29a21596cdf2f95ab3a2f53b876bf5c929e99a).

Changes:
- Do not add `basename` -- it does not seem to be required anywhere
- Remove `mtime` and `filesize` -- these may be sensitive and is probably not a good idea to expose them

Not changed:
- Your PR also actively removes `id` and `Layer`, but these keys are not added by TileServer-GL or underlying libraries and are coming directly from the MBTiles itself. It does not make sense (and might be undesirable) to list all possible keys that can be present and remove them. If you don't want the `Layer` (or any other) key present in the TileJSON, remove it from the MBTiles metadata table.

PR #86, opened 2016-12-05
Implementing baseURL to support non-standard protocol and base routing, fixes #85

### LINKED ISSUE #85: Support for setup behind a (HTTPs) load-balancer
Hey klokantech team, thanks for your continuous great work on the server & osm2vectortiles!

Following issue:

When running the tileserver behind a load balancer, the **utils.getTileUrls** (and all other URL rewriting helpers) are generating invalid URLs.

Currently they get generated by combining ``req.protocol + '://' + req.headers.host``
This is totally fine, as long as the tileserver is the endpoint itself.

In my use case, i got

``https://loadbalancer/foo/[*]`` which get's forwarded to one of the HTTP tile servers ``http://192.168.0.x/[*]``

The problem now: **req.protocol** will be **http** (as lb -> tileserver uses it), **req.headers.host** the cluster-internal hostname of the pool's tileserver which got the request. 
 
Long story short: all paths will be generated for the internal host, with internal non-https, instead of pointing it correctly to the loadbalancer's https endpoint.

Would love to propose a new config option named baseURL, which, when set, is being used instead of ``req.protocol + '://' + req.headers.host``.

PR coming & Cheers!

rastapasta:
see #85

- added another config flag called ``baseURL``
- if ``baseURL`` is set, all paths will be relatived to this base URL (instead of ``req.protocol+host``)
- tested against our https tileserver-gl cluster - yey!

Cheers!

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -66,6 +66,16 @@ The value of ``root`` is used as prefix for all data types.
 
 You can use this to optionally specify on what domains the rendered tiles are accessible. This can be used for basic load-balancing or to bypass browser's limit for the number of connections per domain.
 
+``baseURL``
+-----------
+
+When you run the tile server server behind a load-balancer, the automatic URL generators wont't have the required information to build URLs which point to the correct LB endpoint.
+
+Also, if your load balancer is forwarding a HTTPS connection to the non-HTTPS ``tileserver-gl``, setting the ``baseURL`` is indispensable.
+
+Not compatible with the ``domains`` option.
+
+
 ``formatQuality``
 -----------------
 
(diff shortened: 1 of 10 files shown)


petrsloup:
Closing, see #85

Wykks:
That's too bad, I actually need to serve from non-root path and the PR is exactly what I need.

klokan:
If you need the map tiles only (PNGs and PBFs) you can always implement rewrite of request in nginx/apache/.., which should sit in front of node.js anyway.

Best practice suggests serving the map tiles from multiple sub-domains - so any production deploy will have it's own domain.

PR #93, opened 2016-12-19
Fix: Right code for working with external vector tiles.

ArsenyN:
With this commit code will have right behaviour with external vector tiles.
See: https://github.com/klokantech/tileserver-gl/issues/91#issuecomment-267939138


petrsloup:
@ArsenyN Merged, thanks!

valtorance:
@ArsenyN  - what did your final style.json look like with the external vector tiles?

PR #103, opened 2017-01-09
Use Mapbox GL Inspect for X-Ray View

lukasmartinelli:
I built all the X-Ray functionality into https://github.com/lukasmartinelli/mapbox-gl-inspect.
I am using this from Maputnik as well so I can share all the x-ray/debug functionality.

For tileserver-gl this has the following advantages:
- Show popup when hovering for vector
- Also show polygons in X-Ray (not just outline)
- Light mode (fits better to Tileserver GL because everything is white)
- Show the inspect button on the vector maps as well to directly inspect vector underlying the styles

**X-Ray:**

screenshot from 2017-01-09 14-13-29

**Viewer**:

peek 2017-01-09 14-40


klokan:
Great, thanks for the PR!

I believe, some people want to embed the fullscreen viewer with selected style zoomed to a defined place into websites/blog by iframe - just to show a place or map of given region to their web visitors.
The x-ray/inspect is too technical function for the standard end-user facing viewer - people may get confused in what is happening with the map...

IMHO It should be applied only for the data preview, if the style is not there. But this is my personal opinion only. We can discuss with @petrsloup and @daliborjanak, if this makes sense in the default style viewer in TileServerGL.

BTW did you switch all the colors and name because of the request at https://github.com/maputnik/editor/issues/8#issuecomment-250563805?
The code for x-ray (under this name) is in Mapbox Studio Classic which is 100% open-source code...

lukasmartinelli:
> I believe, some people want to embed the fullscreen viewer with selected style zoomed to a defined place into websites/blog by iframe - just to show a place or map of given region to their web visitors.

Ah alright. If it is meant to embedd it via iFrame it should be removed from the style page!
One can provide a own color function to the inspect plugin http://mapbox-gl-inspect.lukasmartinelli.ch/examples/custom-color-1.html#3/40.00/-74.50
I can use the color function from before - I just didn't know from where it was copied from.
The colors in the default color function have an alpha value which makes overlaying polygons visible.

> BTW did you switch all the colors and name 
> I believe, some people want to embed the fullscreen viewer with selected style zoomed to a defined place into websites/blog by iframe - just to show a place or map of given region to their web visitors.

Ah alright. If it is meant to embedd it via iFrame it should be removed from the style page!
One can provide a own color function to the inspect plugin http://mapbox-gl-inspect.lukasmartinelli.ch/examples/custom-color-1.html#3/40.00/-74.50
I can use the color function from before.

> BTW did you switch all the colors and name because of the request at maputnik/editor#8 (comment)?
The code for x-ray (under this name) is in Mapbox Studio Classic which is 100% open-source code...

If think it's fair to not call it X-Ray since Tom asked so politely. Sometimes it's not all about licensing ;)
I call this mode Inspect in Maputnik - Tangram Play calls it Inspection mode as well.

klokan:
+1 on everything - including the switch of colors and the name from X-Ray to Inspect - shown only for raw data.

The source of the original color palette (only for reference):
https://github.com/mojodna/tilelive-xray
https://github.com/mapbox/tilelive-vector/search?utf8=%E2%9C%93&q=xray&type=Issues

petrsloup:
@lukasmartinelli Look good, thanks! Can you please just remove the last commit from this PR so I can merge this as is?

lukasmartinelli:
> @lukasmartinelli Look good, thanks! Can you please just remove the last commit from this PR so I can merge this as is?

Rebased and removed the commit.

petrsloup:
Thanks

Issue #101, opened 2017-01-05
/fontstacks.json for maputnik


klokan:
The Maputnik editor can load the list of fonts from a fontstacks.json file similar to:

https://github.com/openmaptiles/fonts/blob/gh-pages/fontstacks.json

This can be generated by TileServerGL automatically.

Now it is only served under http://fonts.openmaptiles.org/ and it is a static file.

Next to that, TileServerGL does not serve fonts which are not used in any style (to protect the font files). Maybe implementation of fontstacks.json and removal of the font protection serving can be made together in one commit - as a single config option for public fontstacks.

There is no sponsor to implement this. A clean PR is welcome from the community.

lukasmartinelli:
I don't think anyone from the community will implement this :)

At least the fontstacks.json is important #104 otherwise Maputnik doesn't show any fonts for all the OpenMapTiles styles (they now all switched to the public demo tileserver - question is how ideal is that).

Will fonts.openmaptiles.org still be active? If all styles point to the demo tileserver anyway?

petrsloup:
I did this together with fixing some minor bugs and merging #104 .

@klokan 
It's now possible to add `"serveAllFonts": true` to `options` in config to serve all fontstacks from the fonts directory.

PR #104, opened 2017-01-10
Support fontstacks.json

lukasmartinelli:
Supports fontstacks.json for #101. I don't want to dig into allowed fonts but this at least provides a fontstack.json

### DIFF
--- a/src/serve_font.js
+++ b/src/serve_font.js
@@ -27,6 +27,11 @@ module.exports = function(options, allowedFonts) {
     });
   });
 
+  app.get('/fontstacks.json', function(req, res, next) {
+    res.header('Content-type', 'application/json');
+    return res.send(Object.keys(existingFonts));
+  });
+
   app.get('/:fontstack/:range([\\d]+-[\\d]+).pbf',
       function(req, res, next) {
     var fontstack = decodeURI(req.params.fontstack);


petrsloup:
@lukasmartinelli Thanks for the PR, but I had to do it slightly differently -- listing `existingFonts` is not entirely correct, because these fonts may not be actually accessible (not allowed). See #101

lukasmartinelli:
> @lukasmartinelli Thanks for the PR, but I had to do it slightly differently -- listing existingFonts is not entirely correct, because these fonts may not be actually accessible (not allowed). See #101

Great! Thanks for implementing this Petr!

petrsloup:
@lukasmartinelli Oh, I see there's been some confusion.. I implemented the json to be served on the root: https://free.tilehosting.com/fontstacks.json?key=Og58UhhtiiTaLVlPtPgs together with other "array" jsons (http://tileserver.readthedocs.io/en/latest/endpoints.html#tilejson-arrays). But I believe it's not a problem.. right?

lukasmartinelli:
> @lukasmartinelli Oh, I see there's been some confusion.. I implemented the json to be served on the root: https://free.tilehosting.com/fontstacks.json?key=Og58UhhtiiTaLVlPtPgs together with other "array" jsons (http://tileserver.readthedocs.io/en/latest/endpoints.html#tilejson-arrays). But I believe it's not a problem.. right?

In Maputnik I am infering the URL of the fontstacks JSON from the glyphs URL which for Tileserver GL needs to be `https://free.tilehosting.com/fonts/{fontstack}/{range}.pbf?key={key}`.

Then I cut away the dynamic part `{fontstack}/{range}.pbf` part and ask for the fontstacks JSON there `https://free.tilehosting.com/fonts/fontstacks.json`.

I am not sure if that should be changed to the root from the Maputnik perspective since it makes it more difficult for serving fonts from GitHub (since you would always need a domain).

lukasmartinelli:
What I mean is IMHO tileserver-gl should host the fontstacks.json at `fonts/fontstacks.json` because `fonts` is where the glyphs endpoint is.

petrsloup:
Well, we have:
- `/styles/..` and 
- `/styles.json`
- `/data/...` and 
- `/data.json`

so `/fonts/...` and `/fonts.json` would actually make the most sense

lukasmartinelli:
I need a consistent rule for Maputnik I can apply to the glyphs url - and it should also work when you just throw the font PBFs in GH Pages.

So it would always be the baseurl of the `glyphs` URL + `.json` extension?
That would be ok I guess.

klokan:
+1 on `/fonts.json`.

It makes sense for TileServerGL.

How with hosting on github? 

https://fonts.openmaptiles.org/fontstacks.json
will become https://fonts.openmaptiles.org/fonts.json - you hit root, then "fonts.json" instead of ".json"?

lukasmartinelli:
Having the rule depend on the root domain is not ideal. I still think it is better go relative from where the glyphs are stored.

Example (from OSM Liberty):

Glyphs URL: `https://orangemug.github.io/font-glyphs/glyphs/{fontstack}/{range}.pbf`

If you go to root it would try to request `orangemug.github.io/fonts.json` - an entirely different GitHub page and the process breaks down.

Instead go relative to `https://orangemug.github.io/font-glyphs/glyphs/fonts.json` which allows the user to put the `fonts.json` inside his GitHub page (or whatever other filename - i don't have an opinion whether call it `fonts` or `fontstacks`)

klokan:
I suggest logic: replace `/{fontstack}/{range}.pbf*` with `.json`, unless you are on root of domain already - then use `/fontstacks.json`.

`https://free.tilehosting.com/fonts/{fontstack}/{range}.pbf?key={key}` ->
`https://free.tilehosting.com/fonts.json`

`https://orangemug.github.io/font-glyphs/glyphs/{fontstack}/{range}.pbf` -> `https://orangemug.github.io/font-glyphs/glyphs.json`

`http://fonts.openmaptiles.org/{fontstack}/{range}.pbf` ->
`http://fonts.openmaptiles.org/fontstacks.json`

This is a change on the Maputnik side only, right?

petrsloup:
Changed to `/fonts.json` in the latest release

lukasmartinelli:
Hmm now I made a workaround especially for tileserver GL... so now I will need to change again. https://github.com/maputnik/editor/commit/404b53587f26b54002ed7d759b36698bb8bab342

Please tell me when it is deployed! Because then it will not work anymore until I've updated Maputnik again.

petrsloup:
@lukasmartinelli It's deployed now.

Issue #22, opened 2016-06-24
SIGHUP for reloading configuration


klokan:
It may be practical to gracefully reload config when the server receives a SIGHUP signal - so new maps and styles are used without downtime on the server caused by restarting.

petrsloup:
Implemented in #155

PR #155, opened 2017-04-26
Reload configuration on SIGHUP

tschaub:
I'm opening this PR to hear thoughts on how people think #22 should be implemented.

This brings in [`http-shutdown`](https://www.npmjs.com/package/http-shutdown) for gracefully closing the server before reloading the config.  If there is concern about the extra record keeping for tracking connections, [`server-destroy`](https://www.npmjs.com/package/server-destroy) is slightly lighter alternative, but this would mean a hard shutdown after some fixed timeout.

In order to reload the config and styles, the entire `require.cache` is cleared.  If this is a concern, a more selective cache clearing could work, but I think there would probably be marginal benefit.

I also made it so the `callback` is only called once (previously it was called [on `'listening'`](https://github.com/klokantech/tileserver-gl/blob/49a779970e3443bffaf2c5f58e0e491ab21fb6d3/src/server.js#L363) and [after 1000 ms](https://github.com/klokantech/tileserver-gl/blob/49a779970e3443bffaf2c5f58e0e491ab21fb6d3/src/server.js#L370).  It looks to me like this is never used, but I'm not sure if people are requiring this module directly and relying on the current callback behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,10 @@
     "test": "mocha test/**.js"
   },
   "dependencies": {
-    "async": "2.2.0",
+    "@mapbox/mapbox-gl-native": "3.4.4",
+    "@mapbox/sphericalmercator": "1.0.5",
     "advanced-pool": "0.3.2",
+    "async": "2.2.0",
     "base64url": "2.0.0",
     "canvas": "1.6.5",
     "clone": "2.1.1",
@@ -29,6 +31,7 @@
     "express": "4.15.2",
     "glyph-pbf-composite": "0.0.2",
     "handlebars": "4.0.6",
+    "http-shutdown": "^1.2.0",
     "mbtiles": "0.9.0",
     "morgan": "1.8.1",
     "node-pngquant-native": "1.0.4",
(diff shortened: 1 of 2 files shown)


petrsloup:
Thanks for the PR!

This way, however, the reload is not completely without downtime.

The metadata from .mbtiles is read asynchronously (and parts of the initialization process depend on this). If there are some incoming requests during this small time window, they will probably fail. Plus the server startup itself can take some time.

But solving this absolutely without downtime would require running two separate instances of the server and switching, which seems like an unnecessary overkill. (But it can always be done on the docker level.)

I'm in favor of merging this (unless there are some ideas how to elegantly achieve the configuration reloading with no downtime).

tschaub:
@petrsloup - an alternative would be to do this:

 1. reload all file based config
 2. remove all existing routes and middleware
 3. reconfigure all routes and middleware

The first step is async and the next two are sync.  This could be done without missing any incoming requests.  It would involve a more significant reorganization of how the config/styles are read and how the express app is configured.  I decided not to tackle that here, but think it could make a nice enhancement later.

petrsloup:
I agree these would require more significant reorganization and restructuring.

I'm merging this now.

If anybody has the need to push the configuration reloading to the next level, feel free to open a pull request later.

petrsloup:
@tschaub Thanks for this contribution!

BaptWaels:
Hi,

Thanks for the PR :). I'm using tileserver-gl with the up to date container from the Docker hub (klokantech/tileserver-gl). 

``` 
$ docker run -v $(pwd):/data -p 8080:80 klokantech/tileserver-gl --verbose
```

From my host, I modify the file $(pwd)/config.json. In order to reload my server with the new confs, I'm obliged to restart my container. Any suggestion to prevent this restart ?

PR #156, opened 2017-04-26
Remove unused callback

tschaub:
The `callback` provided to the main server export is called twice on success (once on `'listening'` and once after 1000 ms).  The callback is not being used elsewhere in the repo, so in case it isn't actually required, this pull request proposes removing it.

It's very possible that I'm missing something about how the callback is used.  No offense if you close this :).

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -28,7 +28,7 @@ if (!isLight) {
   serve_rendered = require('./serve_rendered');
 }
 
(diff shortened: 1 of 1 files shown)


petrsloup:
Yes, it is probably just a leftover from when we were setting up tests and does not seem to be used anymore.

Merging, thanks.

PR #170, opened 2017-05-26
Allow node versions >5

robstolarz:


petrsloup:
TileServer-GL v2.0.0 now uses Node 6, so this PR is no longer relevant.

PR #179, opened 2017-06-20
Log errors to stderr and return

tschaub:
This makes it so errors in `serve_rendered.js` are written to `stderr`.  The `return`s are to avoid additional errors when the second arg to the callback is not defined.  It may be that the correct behavior is to reject (at least in [the first case](https://github.com/tschaub/tileserver-gl/blob/de83021c3db53a4b258f119e7f683e51bade02f8/src/serve_rendered.js#L240)) or otherwise handle the error.


petrsloup:
It's definitely better to `return` than to keep failing when these errors happen.

PR #187, opened 2017-06-29
Use debian stretch nodejs repo

aleksejleonov:
Dockerfile based on `debian:stretch`, so better to use native `stretch` nodejs repo instead of `jessie` repo.


petrsloup:
Oh, yes. Thanks for noticing and making the PR!

PR #207, opened 2017-08-22
Try to create the appropriate data type on http errors

### LINKED ISSUE #171: Error: unsupported image type
We're seeing frequent `[Error: unsupported image type]` messages logged after the service has been under load for a number of hours.  Another symptom of a problem is very long response times (30s to 1min) for tiles.  Restarting the service solves the issue for a while.

I'll get some better monitoring in place to try to determine what is going on, but I'm curious if the message provides any clues on where to start looking (I see it is not generated by code in this repo).

Thanks for any tips.

tschaub:
When rendering raster tiles, if a style is configured with a source that returns a non-`200` response, the mbgl callback is [called with `{data: new Buffer(0)}`](https://github.com/klokantech/tileserver-gl/blob/bb09f3df644d5279a83dae990c9e6739ccf97cca/src/serve_rendered.js#L160).  If the source tile format is PNG or JPEG, this results in a message like this from mbgl:

```js
{ 
  class: 'Style',
  severity: 'ERROR',
  text: 'Failed to load tile 3/6/5=>3 for source mosaic: unsupported image type'
}
```

And no response from tileserver-gl.

See https://github.com/klokantech/tileserver-gl/issues/171#issuecomment-323847064 for an example style.  If the source results in anything but a 200 with an image, tileserver-gl logs an error and does not generate a response.

I've put together a fix that returns "empty" PNG, JPEG, or WebP data and maintains the same behavior as before for other URLs (`{data: new Buffer(0)}`).

Let me know if you'd suggest another way to handle this.

Fixes #171.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -37,14 +37,6 @@ mbgl.on('message', function(e) {
 });
 
 /**
- * Cache of response data by sharp output format.  The empty string represents
- * an unknown or unsupported format.
- */
-var cachedErrorResponses = {
-  '': new Buffer(0)
-};
-
-/**
  * Lookup of sharp output formats by file extension.
  */
 var extensionToFormat = {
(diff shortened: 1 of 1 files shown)


petrsloup:
Nice catch, thanks for the PR!

The handling is quite similar to the handling of missing tiles in local mbtiles (https://github.com/klokantech/tileserver-gl/blob/master/src/serve_rendered.js#L271, the `emptyTile` is created for each local mbtiles on init so it is available when rendering).

It would be nice to unify this (the local mbtiles could use the same by-format cache as the remote sources): https://github.com/klokantech/tileserver-gl/blob/master/src/serve_rendered.js#L112.

tschaub:
@petrsloup - I've remove the creation of `source.emptyTile` and reused the same function for creating empty responses for local and remote requests.

petrsloup:
Great, thank you!

PR #209, opened 2017-09-16
ADD support for encoded path

hoenic07:
The current way of sending the path generates a very long URL. An alternative way of providing the path coordinates is by using Google polyline encoding algorithm.

Also the docs for the latlng parameter did not state the value that should be sent.


canadaka:
Why was this not merged? The MapTiler service supports encoded paths with the "enc:" param. So it would be good if tileserver-gl could also support this.

Issue #197, opened 2017-07-25
Strange error with docker container : libEGL warning: DRI3: xcb_connect failed


christopheblin:
Here is a simple script to reproduce the problem :

ubuntu@geoserver:~/tileserver$ ls

=> the directory is empty

ubuntu@geoserver:~/tileserver$ docker run --name tileserver -it -v $(pwd):/data -p 8080:80 klokantech/tileserver-gl
Starting tileserver-gl v2.1.0
Downloading sample data (zurich_switzerland.mbtiles) from https://github.com/klokantech/tileserver-gl/releases/download/v1.3.0/zurich_switzerland.mbtiles
Automatically creating config file for zurich_switzerland.mbtiles
Run with --verbose to see the config file here.
Starting server
Listening at http://[::]:80/
Startup complete

=> everything works fine, using Ctrl+C to exit interactive mode

ubuntu@geoserver:~/tileserver$ docker start -a tileserver
Starting tileserver-gl v2.1.0
No MBTiles specified, using zurich_switzerland.mbtiles
Automatically creating config file for zurich_switzerland.mbtiles
Run with --verbose to see the config file here.
libEGL warning: DRI3: xcb_connect failed
libEGL warning: DRI2: xcb_connect failed
libEGL warning: DRI2: xcb_connect failed
terminate called after throwing an instance of 'std::runtime_error'
  what():  eglInitialize() failed.

/usr/src/app/run.sh: line 8:    11 Aborted                 (core dumped) node /usr/src/app/ -p 80 "$@"
ubuntu@geoserver:~/tileserver$

I have no idea what is going here ...
It seems I can start the container once but the following tries will always crash with this error.

Is this something normal ?

DanielCeregatti:
I'm also getting this error, except not with docker. I've installed tileserver-gl on Debian 9 using npm according to these instructions http://tileserver.readthedocs.io/en/latest/installation.html#npm. I had to add many more packages than the ones listed there before the npm install succeeded, namely:

```apt install pkg-config libjpeg62-turbo-dev libgit-dev libgles2-mesa libegl1-mesa```

Here's what happens when I run tileserver-gl with one of my own mbtiles files:

```
$ tileserver-gl blockgroup.mbtiles
Starting tileserver-gl v2.1.0
Automatically creating config file for blockgroup.mbtiles
WARN: MBTiles not in "openmaptiles" format. Serving raw data only...
Run with --verbose to see the config file here.
[...]

PR #214, opened 2017-09-22
Wait for X server, Xvfb, to come up. Solves Issue #197

polar:
This pull request will properly solve Issue #197: https://github.com/klokantech/tileserver-gl/issues/197
**Strange error with docker container : libEGL warning: DRI3: xcb_connect failed** 

This solves the problem by adding x11-utils to the Dockerfile and using xdpyinfo repeatedly until the Xvfb server is up.

This pull request may have to be changed a bit, in that I've modified the package.json to point to my repository.  Perhaps I should have created a branch? Anyway, you get the general idea.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get -qq update \
     libprotobuf-dev \
     libxxf86vm-dev \
     xvfb \
+    x11-utils \
 && echo "deb https://deb.nodesource.com/node_6.x stretch main" >> /etc/apt/sources.list.d/nodejs.list \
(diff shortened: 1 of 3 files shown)


polar:
I closing this pull request, because an inadvertent typographic error which caused the timeout to be set to 30 seconds, which sometimes isn't enough. Perhaps, the timeout should be parameterized>

petrsloup:
Thanks for the pull request! The general idea made sense.. I think the 30s timeout is not a major concern (if 30s is not enough time, there is probably some other major issue).

PR #222, opened 2017-10-01
Make raster tile renderer pool sizes configurable

otsaloma:
Here's a change to make the raster tile renderer pool sizes configurable in the configuration file. Details of how it works are in the config file diff.

At first I thought for simplicity that I'd use a single value for all scale factors, but I concluded it would have been confusing if the default value depends on the scale factor, but it wouldn't have been configurable at that level. Allowing arrays shorter than `maxScaleFactor` should keep this simple enough to configure.


petrsloup:
Nice and clean PR, thanks!

Issue #223, opened 2017-10-03
Better Support for API Keys


caneroj1:
This is mostly a feature request, as I need better control for replacing API keys in the styles provided by the tileserver.

Currently, it looks like when a request for style JSON comes in, if there is a `key` query parameter, we add that query parameter to sources if the source is served locally.

Would it be possible to be able to add finer control here, and add the `key` query parameter to arbitrary urls that we may specify in the config file? 

I would be willing to work on a pull request for this, as it's something I need in short order.

PR #224, opened 2017-10-03
Add better support for API Keys

caneroj1:
This pull request aims to address issues raised in #223:

1. Adds a new, optional `auth` section to the `options` object in the config file. The "auth" section can contain the following attributes
  a. `keyDomains`: an optional array of external domains to which you want to add API key parameters
  b. `keyName`: the optional (defaults to "key") query parameter name you wish to use to look for API keys in incoming requests and when substituting into urls. For example, "authToken" instead of "key".
  c. `forceSpriteKey`: boolean allowing the user to override the behavior of not passing API key parameters to sprite urls.
2. When serving requests for style JSON, if there is a query parameter corresponding to `keyName`, we add that query parameter to all urls for data served locally, in addition to those whose domains are included in the `keyDomains` list.
3. When serving requests for tile JSON, if there is a query parameter corresponding to `keyName`, we add that query parameter to the urls generated for the tile JSON.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -172,7 +172,7 @@ module.exports = function(options, repo, params, id, styles) {
     info.tiles = utils.getTileUrls(req, info.tiles,
                                    'data/' + id, info.format, {
                                      'pbf': options.pbfAlias
-                                   });
+                                   }, options);
     return res.send(info);
   });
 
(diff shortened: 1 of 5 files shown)


caneroj1:
@petrsloup Could I get some feedback on this pull request? This kind of functionality is something I need at the moment. I'd like to work with you guys to get this back into tileserver-gl proper, so I don't have to work off of a fork.

davidkuehner:
Why is this PR Closed and not merge ? Those options could be really useful !

PR #227, opened 2017-10-20
Wait for XvFB to actually come up.

polar:
We had some issues of the XVFB not coming up. People are upping the time to like 3 seconds. On some installations we have waited much longer. We decide within a specific timeout to wait for the XVFB to some up using xdpyinfo. 

I'm sorry for not squashing my commits, but the changes to the Dockerfile and run.sh are straight forward.


petrsloup:
@polar Thank you for the effort. Could you please clean up this PR so it can be merged? It should only address the single xvfb issue (upgrading packages should not be part of this PR). And also please cleanup (squash) the commit history. You can update this PR by force pushing to your branch.

guenhter:
@polar Commit seems really useful. Would you be so kind and do the cleanup?

petrsloup:
Superseeded by #313

PR #229, opened 2017-10-23
Handle promise rejection

tschaub:
This makes it so promises are more consistently rejected on error and adds promise rejection handling.


petrsloup:
Merged. Thanks!

PR #230, opened 2017-10-24
add public_url option

albanm:
Hello,

I needed to expose the tileserver on a subdirectory behind a reverse proxy. The configuration in https://github.com/klokantech/tileserver-gl/issues/56 was too hackish. So I added an option to explicitly give a public url to tileserver so that it doesn't try to make a guess using headers (still uses current policy when the option is not defined).

I suppose some other users might be interested.

I wrote Dockerfile_test simply to be able to run the unit tests without installing stuff on my OS. But it is not really related to the PR.


petrsloup:
Could you please update this PR ? Most importantly nomnom was replaced with commander in https://github.com/klokantech/tileserver-gl/pull/263

Could you also add a line to the usage documentation (https://github.com/klokantech/tileserver-gl/blob/master/docs/usage.rst) about this feature?

sweco-semhul:
Seeing some issues with sprites requests when using the public_url setting. Not able to see whats the real problem. But I'm guessing it could be related to fixUrl method for paths not specified with local:// but only with the server path.

lamuertepeluda:
any news on this? I think it'd be a pretty useful feature...

richturner:
This requirement has come up in several places and a PR was submitted in #86 but still this functionality isn't finding its way into the code base....here's another vote for this feature as it really is useful.

richturner:
If you want to use this `public_url` functionality in a docker image I've created a `Dockerfile` that pulls the source files from this fork just build this image:

https://gist.github.com/richturner/25a63f250a4d1f2b1d9883a8e64a0705

petrsloup:
Looking forward for merging this once it's cleaned up against the current master. Looks like a lot of people could benefit from this feature.

albanm:
Sorry about that... I got lazy on this one.

Should be ok after the last commit.

petrsloup:
Thanks !

PR #232, opened 2017-11-03
Catch incorrect header exceptions

rani-pinchuk:
A tile in our data is corrupted. As a result its header is not correct, and this causes the server to crash. We know which exact tile is broken in our data, so we can reproduce the problem.

```
zlib.js:560
throw error;
^
Error: incorrect header check
at Zlib._handle.onerror (zlib.js:370:17)
at Unzip.Zlib._processChunk (zlib.js:550:30)
at zlibBufferSync (zlib.js:239:17)
at Object.exports.unzipSync (zlib.js:151:10)
at /usr/src/app/src/serve_rendered.js:185:38
at Statement.<anonymous> (/usr/src/app/node_modules/@mapbox/mbtiles/lib/mbtiles.js:166:20)
Segmentation fault (core dumped)`
```

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -192,7 +192,7 @@ module.exports = function(options, repo, params, id, dataResolver) {
                   response.data = zlib.unzipSync(data);
                 }
                 catch (err) {
-                  console.log("Skipping incorrect header for tile mbtiles://openmaptiles/%z/%x/%y.pbf", z, x, y);
+                  console.log("Skipping incorrect header for tile mbtiles://%s/%s/%s/%s.pbf", id, z, x, y);
                 }
                 if (options.dataDecoratorFunc) {
                   response.data = options.dataDecoratorFunc(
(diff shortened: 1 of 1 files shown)


petrsloup:
I believe we can merge this patch to make the server more robust.

But can you please improve the log message so that it does not contain the hardcoded `openmaptiles` identifier? The id of the data source is in the `id` variable so you can use that.

rani-pinchuk:
mmm... fixed it, but for some strange reason the check failed. However, I suspect that the failing is not related to the one line I have changed. In the log it is mentioned that "Some index files failed to download...", an error about "/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found" and a "TypeError: Cannot read property 'close' of undefined". Non of those are related to the file changed. 

Can you please check that?

petrsloup:
@rani-pinchuk That was just some temporary Travis issue, restarting the build fixed the issue.

Thanks for the PR, merging.

PR #242, opened 2017-11-30
Addresses issue #197, send SIGTERM through to node

samizdis:
Node can't be PID 1, else it can't be term'ed and xvfb won't stop nicely.  So, according to https://docs.docker.com/engine/reference/builder/#entrypoint, we need to make sure it's not the bare entrypoint.

Also we need to be sure that run.sh passes the signal through to node, see https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ENV NODE_ENV="production"
 VOLUME /data
 WORKDIR /data
 EXPOSE 80
-ENTRYPOINT ["/usr/src/app/run.sh"]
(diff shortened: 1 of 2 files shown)


samizdis:
See also https://github.com/nodejs/node-v0.x-archive/issues/9131

petrsloup:
Thanks for the PR! I agree this could help, although the entrypoint change was not needed (there's `#!/bin/bash`).

PR #263, opened 2018-03-08
Replace Nomnom with Commander

loicgasser:
Fix for https://github.com/klokantech/tileserver-gl/issues/257

If you're wondering why I use version 2.1.0 of commander:
https://github.com/tj/commander.js/pull/757

I saw that the options `--cors` defaults to true.
I thought it would make more sense to disable CORS on demand via `--no-cors`.

I also don't use the `.version()` of commander since `-V` is already used for the `verbose`.


loicgasser:
I can actually overwrite the flags of `version(str, flags)`, so it's fixed.
For the positional argument, is should be done as well, but do you want to also check the `.mbtiles` extension?

petrsloup:
The `.mbtiles` extension checking is not needed. It's just important to keep the compatibility with the previous version.

Thanks.

PR #269, opened 2018-03-19
Silent mode (do not log requests with status == 200)

michaill:
Adding command option -s, --silent to stop logging successful requests.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -46,6 +46,10 @@ var opts = require('commander')
     '-V, --verbose',
     'More verbose output'
   )
+  .option(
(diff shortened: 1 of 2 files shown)


petrsloup:
Nice idea! Would it make sense to also skip 304 reponses?

michaill:
I am glad you like it. My motivation was to minimize log output in our production and 304 responses never occur there. It is definitely good idea to add them, too. So here it is :)

petrsloup:
Thanks!

jxlarrea:
204 should also be added to this!

PR #270, opened 2018-03-22
Add silent to doc

loicgasser:
I just saw:
https://github.com/klokantech/tileserver-gl/pull/269

this is a simple doc fix.

I take the chance to ask you if there is any intreset in adding a `/styles/` endpoint to tileserver-gl.
I ask the question because I think it could simplify the integration with `maputnik`.

What do you think?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -18,6 +18,7 @@ Getting started
     -p, --port <port>     Port [8080]
     -C|--no-cors          Disable Cross-origin resource sharing headers
     -V, --verbose         More verbose output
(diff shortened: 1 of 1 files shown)


loicgasser:
ping @petrsloup

petrsloup:
Thanks for the PR.

BTW There is already `/styles.json` endpoint listing the styles + HTML list of the styles and data at the root. I don't think anything more is needed.

loicgasser:
Ah thanks yes I missed that.

PR #273, opened 2018-04-03
#119 - Fixes support for proxy (X-Forwarded-Host header).

### LINKED ISSUE #119: X-Forwarded-Host Not Used in URLs
I set up a tileserver-gl-light instance behind an IIS reverse proxy using IIS Rewrite. I set the X-Forwarded-Host request header appropriately. However, the URLs that are returned in my styles do not have the hostname from the X-Forwarded-Host header. The URLs have the hostname from the Host header.

The documentation for the express library says to use req.hostname, but tileserver-gl is using req.headers.host.

danpe:
Fixes #119 

https://expressjs.com/en/guide/behind-proxies.html


petrsloup:
Thanks for the PR.

I agree it's the correct solution to let the express handle `X-Forwarded-Host`.

Unfortunately, there is a bug in Express v4 and `req.hostname` and `req.host` do not contain port information. (So this PR would break all instances running on a port different than 80 -- including local development and probably some production deployments.)

This will be fixed in Express v5 (see https://expressjs.com/en/guide/migrating-5.html#req.host), but we should definitely wait for the stable release before migrating.

danpe:
Thanks Petr, 

Do you think maybe it's worth combining `req.hostname` with `req.headers.host` to maintain the port as a workaround until Express 5 is released?

Let me know,

petrsloup:
Although it should be possible to parse the correct value from the `req.hostname` and `req.headers.host` it's kind of a hack. And it is actually not a correct solution in some cases, since the proxy might be running on a different port than the backend server (this seems to be a little messy at the moment -- `X-Forwarded-Host` usually does not contain port information and people sometimes use `X-Forwarded-Port`, but I'm not sure if express will respect this).

Anyway, the development on this is actually not a priority for our team at the moment -- our CDN and proxies do not modify the Host header so no additional handling is required.

I suggest you do the same before we can merge this PR (after Express v5 is stable) -- I think it's the best solution at the moment.
I don't know what proxy you are using, but see for example https://www.ctrl.blog/entry/how-to-httpd-reverseproxy-hostheader and especially https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost.

danpe:
I made a nice workaround:
```
app.all('*', function (req, res, next) {
    req.resolvedHost = req.headers['x-forwarded-host'] ? req.headers['x-forwarded-host'] : req.headers.host;
    next();
});
```

Then I access the `resolvedHost`, you want me to update the PR with that fix? or you think I should combine `X-Forwarded-Host` with `X-Forwarded-Port` first?

danpe:
@petrsloup  Is this solved with `publicUrl`?

jfgaudreault-p:
As I understand it X-Forwarded-Host is still not fixed or working?
The documentation is a bit misleading: https://github.com/maptiler/tileserver-gl/blob/master/docs/deployment.rst
I thought this used to work when I was using an older github repo of an earlier tileserver-gl version (2.x?)

It says it is important to set X-Forwarded-Host but it seems to have no effect when getting a style.json for example.
X-Forwarded-Proto works fine on the other side.

Issue #46, opened 2016-08-29
No tile found should return 204 instead of 404


klokan:
Related to:
mapbox/mapbox-gl-js#1800 (comment)
and
https://github.com/klokantech/tileserver-php/pull/88

amirmasud:
Hi Guys!
when would it be fixed guys?  is there any plan?

petrsloup:
Done in #278

PR #278, opened 2018-05-07
Return 204 when mbtiles error is 'Tile does not exist'

amirmasud:
Result status code changed from 404 to 204 when mbtiles error is 'Tile does not exist'  (due to issue #46 )


petrsloup:
Thanks!

PR #303, opened 2018-08-03
Update commander and devDeps

loicgasser:
Default value and coercion order are now inverted.

This seems to be the new undocumented behavior of commander. See: https://github.com/tj/commander.js/pull/829

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "canvas": "1.6.8",
     "clone": "2.1.1",
     "color": "1.0.3",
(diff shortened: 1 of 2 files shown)


loicgasser:
@petrsloup any reason not to merge this?

petrsloup:
Thanks! Just to clarify -- this does not fix any bug, but it allows us to update the commander dependency.

loicgasser:
Yes exactly

Issue #274, opened 2018-04-07
WMTS support without using wmts.maptiler.com


spatialillusions:
It would be really nice if it was possible to get WMTS support without using wmts.maptiler.com. The depencency on using wmts.maptiler.com makes it quite tricky to use in a network not connected to the internet.
I might be able to look into creating a pull request for this if you think that it might be a useable feature.

blairdgeo:
:+1:

spatialillusions:
Now we are just waiting for pull #307 to be merged for this to be implemented.

petrsloup:
Implemented in #307, thanks!

PR #307, opened 2018-08-17
Add support for WMTS capabilities

spatialillusions:
This will implement #274 and add support for WMTS Capabilites without needing to connect to wmts.maptiler.com. I have tested the code in QGIS to verify that I am able to connect to the services. 

Please provide feedback if it is anything you want me to change.


spatialillusions:
Sorry for the outdated template, I did my development in a virtual machine and forgot to move out the latest version of the template. Should be correct now.

spatialillusions:
Would be very nice if @klokan or @petrsloup could give me a timeframe for when/if this will be merged. Thanks!

src/server.js
```diff
@@ -325,11 +326,6 @@ function start(opts) {
               Math.floor(centerPx[1] / 256) + '.' + data_.format;
         }
 
-        var query = req.query.key ? ('?key=' + req.query.key) : '';
-        data_.wmts_link = 'http://wmts.maptiler.com/' +
-          base64url('http://' + req.headers.host +
```
petrsloup: Was this the only reference to the `base64url` package? I guess we could remove the dependency completely...

src/server.js
```diff
@@ -374,6 +370,20 @@ function start(opts) {
     return res.redirect(301, '/styles/' + req.params.id + '/');
   });
   */
+  serveTemplate('/styles/:id/wmts.xml', 'wmts', function(req) {
+    var id = req.params.id;
+    var wmts = clone((config.styles || {})[id]);
+    if (!wmts) {
+      return null;
+    }
+    if(wmts.hasOwnProperty("serve_rendered")&!wmts.serve_rendered){
```
petrsloup: Please fix code formatting here (space after `if`, `&` -> ` && ` (with spaces around), `){` -> `) {`, indentation of the return below, ...)

spatialillusions:
@petrsloup Should I remove base64url from package.json as well? As far as I can tell it is not used in any other places in the code.

petrsloup:
Yes, please remove the dependency completely -- it's not used anywhere else in the project.

petrsloup:
@spatialillusions https://github.com/klokantech/tileserver-gl/blob/e8b3c0254678f049bc1ef67bc44cd5add4aa81da/src/server.js#L300 This should be probably also removed, right?

spatialillusions:
Removed the unused variables, and also added the key to the template so that it should work for your hosted services. (Not sure if the key is used for any other links in the template.)

petrsloup:
Thanks, I don't see any more issues and everything seems to be working now. Merging...

PR #310, opened 2018-09-10
Ensure that container can be restarted

guenhter:
When the file `/tmp/.X99-lock` is available on startup of `Xvfb`, the process does not start.
So the script basically deletes this on startup. 
Usually, this file should be cleaned by the `Xfvb` process itself, that's why the `--retry` is specified. This waits for the process to exit for the defined period of time.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -7,6 +7,8 @@ _term() {
 
 trap _term TERM
 
+# Delete files if they were not cleaned by last run
+rm -rf /tmp/.X11-unix /tmp/.X99-lock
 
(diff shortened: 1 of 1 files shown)


guenhter:
Great to see this merged.
When can we expect a new Docker-Tag in the docker registry (except the latest)?

KleggerKoder:
I encountered this problem yesterday using the command line npm install on Debian 9. 
Had to use following command before calling tileserver-gl to get it working
`Xvfb :1 -screen 0 1024x768x24 </dev/null & export DISPLAY=":1"`

Is this fixed now on NPM?

PR #313, opened 2018-09-14
Wait for Xvfb to startup

guenhter:
Poll displaoy status periodically to see if Xvfb is started instead of just sleeping three seconds.
Supersedes #227


spatialillusions:
I have also had the problem that my container won't restart so if this solves it it would be really nice. The fail to restart occurs when I run docker with `--restart always` and then don't power down my machine gracefully (just kill the power instead of sending a shutdown signal). Then when my machine starts up the container restarts over and over again with the errors similar as mentioned in this case. 
(Of course I always shut down my machines gracefully, but you know, users...)

guenhter:
I had the same problem with `--restart`  but this is not solved by this PR. 
Luckily for you, that problem is already fixed with #310.
Also a new docker tag is already available in the public registry.

This PR concentrates on the issue, that 3 seconds sleep is not needed (because `Xvfb` usually starts up faster) AND when `Xvfb` isn't starting up at all to raise an error already before starting the actual tileserver.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:6-stretch
+FROM node:6.14.4-stretch
```
petrsloup: Why is the node version fixed here? Can't we just use `6` or `6.15` to get the security updates?

guenhter: I usually tend to fix versions to avoid that the "generic" image `6` is downloaed one time and never updated (as long nobody performs a `docker pull` on it). In that case the version `6` could even be older as the pinned version.
What do you think?

petrsloup: I understand the motivation and I'm not strongly opposed to fixing the version, but I think we could fix just the major version and let the minor version to update (`6.15`) or at least update to the latest version (`6.15.1`) ?

guenhter: Alright. Updated the version to the latest one.

PR #337, opened 2018-12-26
Correctly retrieve the public_url parameter on start

mcolmant:
The `public_url` option is not correctly retrieved and make the program crash.
This PR hence fixes #230.


petrsloup:
Thanks!

PR #355, opened 2019-04-13
let the user define a log file or a log format

arichnad:
let the user define a log file or a log format:

log file defaults to standard out.

log format defaults to current behavior ('tiny' for production, 'dev' for dev, and not enabled in test)


petrsloup:
Thanks! Merged.

PR #375, opened 2019-07-02
Mode 11589 - Updating the maps

dobrevkalm:
If you can upload this to the test server. Basically, I've only changed the mono.json (the default style).

The layers position in the json file affects their overlapping.

Some streets and street names show differently simply because the layer itself becomes visible after certain zoom level. Same goes for the landcover rendering.

There are a lot more road labels added (E 47) so I had to reduce the amount showing on screen and increase the zoom from 8 to 9 (road_motorway_shield and road_shield).

I've tried to fix the names. Basically, the names of the countries, oceans, cities, etc. are in English and the point locations and street names are in latin. Changing my browser locale did not in any way affect the way the names appear.


petrsloup:
I think you are creating the Pull request to the wrong repository :-)

You probably want to create PR to https://github.com/monosolutions/tileserver-gl/tree/master (fork of this official tileserver-gl). Then there's only two-commit difference.

PR #378, opened 2019-08-01
Added JWT support, updated a couple of dependencies

Sinnohd:
Hi,

I added support for JWT in the frontend (we use it with Lambda@Edge on AWS). I removed the legacy mapbox.js part and using now only mapbox-gl for both, vector and raster, tiles. Also updated quite a few dependencies (biggest change was the canvas upgrade, however on this one, I didn't found the time to test if the watermark feature is still fully functional). 

Br,
Jens

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "color": "4.2.3",
         "commander": "12.0.0",
         "cors": "2.8.5",
-        "express": "4.18.3",
+        "express": "4.19.1",
         "handlebars": "4.7.8",
         "http-shutdown": "1.2.2",
         "morgan": "1.10.0",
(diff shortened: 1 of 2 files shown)


petrsloup:
Thanks for the PR!

However, it would be nice to have these a separate changes rather than 1 huge batch.

- The JWT support seems to be pretty application-specific and I don't think it should be part of the tileserver-gl.
- The mapbox.js is used for displaying raster tiles mainly for browsers without WebGL support (that's actually the main reason why tileserver-gl even bothers with the server-side rasterization) so replacing it with mapbox-gl kind of defeats the purpose

- The canvas upgrade part is nice and I'd love to see this as a separate PR. Could you find time to create a clean canvas-upgrade-only PR for review?

Issue #344, opened 2019-02-01
text cropped on raster tile


germsb:
Docker image klokantech/tileserver-gl

This problem has already been reported here: #184 #35 [224](https://github.com/openmaptiles/openmaptiles/issues/224) and many other

Apparently the problem came from mapbox-gl-native [bug](https://github.com/mapbox/mapbox-gl-native/issues/11838) but it has been fixed since version 4.X
so finally not so sure that the problem comes from there since tileserver-gl uses this latest version!

As workaround, I have tested `"symbol-avoid-edges": true,` in layers layout without positive effects.
### vector tiles ok
![Imgur](https://i.imgur.com/fSu5PRr.jpg)
### raster tiles
![Imgur](https://i.imgur.com/jOR5UXD.jpg)
![Imgur](https://i.imgur.com/mhuBNAr.jpg)

Whatever the style, the problem persists (even those proposed by default)

I use data from openmaptiles.com custom extract who use the schema v3.9. ( ̶r̶a̶s̶t̶e̶r̶ ̶r̶e̶n̶d̶e̶r̶ ̶i̶s̶ ̶o̶k̶ ̶w̶i̶t̶h̶ ̶k̶l̶o̶k̶a̶n̶t̶e̶c̶h̶/̶o̶p̶e̶n̶m̶a̶p̶t̶i̶l̶e̶s̶-̶s̶e̶r̶v̶e̶r̶) and tested also with extractions done with quickstart.sh from openmaptiles/openmaptiles

So what to do?

Otherwise, very good job, I find it just a pity that a bug that is more than 2 years old and which has a rather heavy impact is not resolved or simply better documented

BHudsonJSI:
I can confirm I see this same issue in tileserver-gl 2.5.0 and osm-2018-12-03-v3.9.

Check Australia at level 4-7 using either osm-bright or klokantech-basic in raster.  Neither style exhibits this issue in vector.

![image](https://user-images.githubusercontent.com/28735994/52588669-02eb3e00-2e0b-11e9-8d09-80a199b30733.png)

![image](https://user-images.githubusercontent.com/28735994/52588729-244c2a00-2e0b-11e9-82d1-22a08d8874f6.png)

![image](https://user-images.githubusercontent.com/28735994/52588838-6d03e300-2e0b-11e9-8bcf-9de12f54bed7.png)

[...]

Issue #184, opened 2017-06-26
Text is cropped on raster tile 


Duke74:
I noticed that when using raster tiles, there is some case where text is cropped.
Here is an example with a peak name situated near the edge of a tile.

![chrome_2017-06-26_18-29-03](https://user-images.githubusercontent.com/1609064/27549902-e50008cc-5a9d-11e7-8860-431587edbb5b.jpg)

Is there any existing workaround to this issue ?

petrsloup:
Duplicate, see #35, #161 and openmaptiles/openmaptiles#191

Issue #35, opened 2016-08-04
Broken map labeling spanning multiple raster tiles


procogent:
Map label elements that span multiple raster tiles are frequently cut off (only appearing on one of the tiles) resulting in odd artifacts.

![z10 example](https://cloud.githubusercontent.com/assets/20845025/17414566/c3368b30-5a3a-11e6-925d-faa3496ccdf4.png)
![z11 example](https://cloud.githubusercontent.com/assets/20845025/17414567/c33755b0-5a3a-11e6-8f7d-aeb14a891b6a.png)

klokan:
Acknowledged. We will discuss how to increase the buffers around tiles to avoid this glitches.

klokan:
We start to work actively on this issue. Upgrade of Mapbox GL Native did not solve the problem.
The next step is digging into the combination of input data (vector tiles + JSON style) on the sample area and debug tests on the GL Native.

petrsloup:
I did some testing and it's possible there are actually two separate issues:
![image](https://cloud.githubusercontent.com/assets/707817/18049518/10d61092-6de9-11e6-86d0-594087433912.png)

- Sometimes, two labels are present very close to each other in both tiles, but it looks like different label is chosen to be used in different tile (blue circles).
- And sometimes the label is present only in one of the tiles and not the other (blue circle).

Since this issues is also happening when using mapbox-gl-native directly (center+zoom+dimensions), there is probably nothing we can do on tileserver-gl side -- it's either bug in mapbox-gl-native or the data.

We will look into the tiles data to better understand what exactly is going on.

klokan:
Related PR: https://github.com/mapbox/mapbox-gl-native/pull/3357

klokan:
@petrsloup - config for "MapMode::Still"?

mikemorris:
This appears like you're hitting https://github.com/mapbox/mapbox-gl-native/issues/3582, which can be fixed by adding `symbol-avoid-edges: true` to the clipped label layer in the style. We're [considering making this default behavior](https://github.com/mapbox/mapbox-gl-style-spec/issues/287), would appreciate any feedback you have, or edge cases that this fails to address.

petrsloup:
This indeed seems to help. Adding `"symbol-avoid-edges": true` to the affected layer ensures the labels are not rendered when touching the tile edge:
[...]

PR #379, opened 2019-08-09
Add tileMargin option

stefslon:
The issue of cropped labels in raster tiles has been reported in multiple times: #344, #184, #35 #224, etc. I have tried a variety of offered suggestions to resolve the issue, but none worked reliably. 

I ended up solving the problem by adding code to generate larger tiles and then crop them to the right size. This can be controlled via a new `"tileMargin"` option in configuration file.

Maybe this will be useful to someone else.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -27,7 +27,8 @@ Example::
       "maxSize": 2048,
       "pbfAlias": "pbf",
       "serveAllFonts": false,
-      "serveStaticMaps": true
+      "serveStaticMaps": true,
+      "tileMargin": 0
     },
     "styles": {
       "basic": {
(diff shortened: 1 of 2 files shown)


petrsloup:
Thanks for the PR!

This is a nicely written "workaround" for the buffering issue, however I'm quite hesitant about merging this into master.

There's potentially huge performance degradation -- the way the tiling works, there is always 1 vector tile needed to render 1 raster tile, but with this option, you suddenly need to load 9 tiles from the .mbtiles (basically 9x the number of sources in the style).

It should be safe (since the default value is 0), but I believe the performance drawback should be added to the documentation (as well as the description of the option itself). Could you please add it to this PR?

stefslon:
Thanks for the feedback. I've added ``"tileMargin"`` description and perfomance drawback notes to the documentation.

joaokho:
> tileMargin

where to put this "tileMargin" setting?

petrsloup:
Thanks, merged.

PR #398, opened 2019-12-02
Lsiden/issues/395

lsiden:
resolves maptiler/tileserver-gl#395

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -387,7 +387,7 @@ module.exports = function(options, repo, params, id, publicUrl, dataResolver) {
 
   var respondImage = function(z, lon, lat, bearing, pitch,
                               width, height, scale, format, res, next,
-                              opt_overlay, opt_bbox) {
+                              opt_overlay) {
     if (Math.abs(lon) > 180 || Math.abs(lat) > 85.06 ||
         lon != lon || lat != lat) {
       return res.status(400).send('Invalid center');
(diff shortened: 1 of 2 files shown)


src/serve_rendered.js
```diff
@@ -481,8 +481,12 @@ module.exports = function(options, repo, params, id, publicUrl, dataResolver) {
           if (!buffer) {
             return res.status(404).send('Not found');
           }
-
+          var hdr_center = 'lat=$lat,lon=$lon'
+            .replace('$lat', lat)
+            .replace('$lon', lon);
           res.set({
+            'tileserver-Center': hdr_center,
+            'tileserver-Zoom': z,
```
robert-claypool: I'm used to seeing headers title cased like `Content-Type` and `X-Forwarded-For`, not `content-Type` or `x-Forwarded-For`. For consistency, maybe use `TileServer-Center` and `TileServer-Zoom`.

RFC now recommends dropping the `X-` prefix, but for application specific headers that will never graduate to web standards, the `X-` is a good indicator it's custom. https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions

lsiden:
I decided I no longer needed this for my project so I dropped it.
Nevertheless, if you think this might be useful to others, I can update the
PR with your suggestions.  Let me know if you are still interested.

-Larry Siden

petrsloup:
I think this is a really specific use case and I'm not going to merge it now.

PR #403, opened 2019-12-21
Upgrade to node10, latest dependencies

nyurik:
use node10 (latest node that has binary built @mapbox/mapbox-gl-native v5.0.2)
use latest canvas and sharp API


petrsloup:
Merging, thanks! This means that the `master` is now becoming the next major version -- `3.0.0`.

PR #404, opened 2019-12-21
Update code to ES6

nyurik:
This should be merged after #403 

* `var`  ->  `let` / `const`
* `[].forEach`  ->  `for (... of ...)`
* `'...' + var`  ->  template strings  \`...${var}\`
* function  ->  arrow functions `=>`
* use `===` and `!==`  instead of `==` and `!=`

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -2,20 +2,20 @@
 
 'use strict';
 
-var fs = require('fs'),
-    path = require('path'),
-    request = require('request');
+const fs = require('fs');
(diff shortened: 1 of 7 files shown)


petrsloup:
Thanks, I'll probably merge this today (after testing out a few things). Did you use some automated tool to do these updates? (as a reference for the future)

nyurik:
@petrsloup most of these were suggested by IntelliJ IDEA

PR #410, opened 2020-01-17
Relax Node.js compatibility for -light version

jdesboeufs:
Hello!

`tileserver-gl-light` works perfectly with Node.js 12 and 13.
This PR relax the field `engines.node` in its `package.json` file.

We actually use a `tileserver-gl-light` fork in French governement digital services, with this improvement :)

Jerome

### DIFF
--- a/publish.js
+++ b/publish.js
@@ -28,6 +28,8 @@ delete packageJson.dependencies['sharp'];
 delete packageJson.optionalDependencies;
 delete packageJson.devDependencies;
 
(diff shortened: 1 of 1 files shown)


klokan:
Thanks @jdesboeufs for the contribution !

petrsloup:
Yes, you are right --  the `-light` version does not have any native dependencies, so the version restriction is unnecessarily strict here. Thanks for the PR!

Issue #350, opened 2019-03-06
Semi-transparent outlines on raw, premultiplied input


justewg:
Thanks for great work, very usefull.
My business task for using tile-server is to show on leaflet/mapbox map three map layers, the lowest is land with some more layers, middle layer is our application data and top layer should contain roads, poi, labels, water regions and should be transparent, of course.
And I have to use raster map mode. So, now my top layer's tiles PNG-images enclose a gray or black substrate under the translucent white color. Result on the attached screenshot.

![image](https://user-images.githubusercontent.com/16482115/53876357-a8af5880-4017-11e9-8c7a-300149784fbc.png)

I found that raster files are producing by sharp-library and post the issue to author:
https://github.com/lovell/sharp/issues/1599

Our common investigations point me to the follow: data, passed to sharp-constructor are premultiplied in mapbox-gl-native library.
But I haven't found the solution, where and how I can prevent it.

May be you can help me?

cgalli:
Hey justewg,

Had a similar issue and was able to back out the alpha channel in serve_rendered.js just before the call to sharp. Seems to work. I applied this js to the data object first:

for (var i = 0; i < data.length; i += 4) {
    var alpha = data[i + 3];
    var norm = alpha / 255;
    if (alpha === 0) {
        data[i] = 0;
        data[i + 1] = 0;
        data[i + 2] = 0;
    } else {
        data[i] = data[i] / norm;
        data[i + 1] = data[i + 1] / norm;
        data[i + 2] = data[i + 2] / norm;
    }
}

Admittedly inspired by https://github.com/consbio/mbgl-renderer/commit/e64af7c3a1b4c97188031caf68f904c6ad749aab after following the issue around. Perhaps that'll help your project get over the hump ours had.

justewg:
That works great. Thanks a lot, @cgalli

xabbu42:
Perhaps a pull request could make it easier to get this fixed in the main repo? Or are there other reasons for the delays?

PR #415, opened 2020-01-28
FIx issue #350 Semi-transparent outlines on raw, premultiplied input

xabbu42:
I'm not sure if there is any other reason why this longstanding issue is not resolved yet. But I figured a pull request with a properly indented fix could help, so here it is. This fixes issue #350.


petrsloup:
Thanks, merged... I was worried a little about the performance, but there does not seem to be significant impact.

PR #416, opened 2020-02-14
update glyph-pbf-composite

tbicr:
glyph-pbf-composite v0.0.2 were released without license that can be issue of licensing complience
as license were added in v0.0.3 NPM release it nice to be updated to licensed software version


petrsloup:
Thanks!

PR #419, opened 2020-02-25
Updated usage.rst to reflect public_url parameter

vipyoung:
Added missing documentation line for public_url.


vipyoung:
Added one line for the public_url parameter

petrsloup:
Thanks, this was an oversight.

PR #433, opened 2020-04-03
Add id to tilejson

frodrigo:
Add the `id` field to tilejson output. So be able to retrieve the tilejson and rebuild the path to direct tilejson for further usage.

* On rendered it's on a add.
* On data it may override the id from the mbtiles.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -145,6 +145,7 @@ module.exports = {
           Object.assign(tileJSON, info);
 
           tileJSON['tilejson'] = '2.0.0';
+          tileJSON['id'] = id;
           delete tileJSON['filesize'];
           delete tileJSON['mtime'];
           delete tileJSON['scheme'];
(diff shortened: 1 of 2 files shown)


klokan:
Could you, please 🙏 , (ideally before making a pull request proposing a change in the software) make an issue explaining:
- what is the pain (=what is not possible to do right now with the software)
- why do you propose to do something (=reason, thoughts context)
- if you propose a solution (such as "Add id to tilejson") also what are all the advantages/benefits and perhaps disadvantages of your proposed change.

Then somebody can reasonably decide whether a pull request is meaningful and improves the software or not.

frodrigo:
First I made code because I need it and plan to use it. Since I think it may benefit to other I share it as PR (and change is very small here). I personally think code is simpler to understand than long text explanation when diff is small.

So, actually `/styles.json` array share style `id`, but data and rendered (/index.json, /rendered.json; /data.json) are available as tilejson, without `id`.

I just need style, renderer and data id to store else where and to point later to unitary tilejson path. I need this in my UI. I also need it in my tile expirer to simply known the tile path and expire. To do it I get the `id` I added to the tilejson arrays.

PR #454, opened 2020-06-06
Remove max node version restriction

autarc:
# Pull Request

The module is currently restricting the node version through an upper bound of `11.X`. Since the [the latest LTS version is 12.X](https://nodejs.org/en/blog/release/v12.13.0) and works fine with the code, this PR removes the the max. node version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "license": "BSD-2-Clause",
   "engines": {
-    "node": ">=10 <11"
+    "node": ">=10"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


gmaclennan:
`@mapbox/mapbox-gl-native` [does not work with Node 12](https://github.com/mapbox/mapbox-gl-native/issues/16485) as far as I understand, and tileserver-gl depends on this for rendering vector tiles to raster tiles.

petrsloup:
@gmaclennan is right. This is why the restriction is in place. tileserver-gl-light package does not have this restriction.

PR #457, opened 2020-06-18
Streamline Dockerfile for caching and size

joakimfors:
Use newer base images and modify Dockerfile to allow for better caching and smaller image file size. Simplify entry script and run as a non-root user.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,6 @@
-.git
-docs/_build
-node_modules
-test_data
-light
-config.json
-*.mbtiles
+*
+!src
+!public
+!package.json
+!package-lock.json
+!run.sh
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,4 +3,4 @@
 !public
 !package.json
 !package-lock.json
-!run.sh
+!docker-entrypoint.sh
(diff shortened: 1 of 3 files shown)


scara:
Cool!
If _multi-stage_ cannot be adopted here (Docker 17.05+), _multi-stage_ could be emulated via https://docs.docker.com/develop/develop-images/multistage-build/#before-multi-stage-builds.

HTH,
Matteo

petrsloup:
Cool, but does the simplified script cover all the cases why the original run.sh is so complicated?

There were some concerns especially about the restartability of the container and waiting for xvfb to start up properly.

See https://github.com/maptiler/tileserver-gl/pull/310, https://github.com/maptiler/tileserver-gl/pull/313 and also  https://github.com/maptiler/tileserver-gl/pull/242

joakimfors:
The xvfb-run wrapper script that is included in the Debian xvfb packages takes care of all setup. However, #242 could be a separate issue. Will check that.

joakimfors:
OK, 6b3f557 should cover the same issue as #242

petrsloup:
Thanks, merged.

Issue #461, opened 2020-07-01
Reflected XSS vulnerability in the application main page


joohoi:
Hi, I stumbled upon a reflected XSS in the software.

It stems from reflecting unsanitized user controller input directly to the HTTP response.

I initially observed this in Tileserver GL 2.5, but from the looks of it it's still present in the current `master`.

Vulnerable request example:
```
https://example.org/?key=%22%3E%3Cscript%3Ealert%281%29%3B%3C%2Fscript%3E
```

The code in question is here https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L322-L323

PR #463, opened 2020-07-02
Fix reflected XSS in 'style' parameter

korpd:
I found another reflected XSS issue which is very similar to one in #461

Example link: https://example.com/?style=%22%3E%3Cscript%3Ealert(2)%3B%3C%2Fscript%3E


tomasgor:
Hi @petrsloup can you publish new package version with this change?

PR #484, opened 2020-10-22
add api key to sprite ressources in style json

candux:
this make some setups easier becouse sprites can be handeld
like any other ressources

### DIFF
--- a/src/serve_style.js
+++ b/src/serve_style.js
@@ -43,7 +43,7 @@ module.exports = {
       }
       // mapbox-gl-js viewer cannot handle sprite urls with query
       if (styleJSON_.sprite) {
-        styleJSON_.sprite = fixUrl(req, styleJSON_.sprite, item.publicUrl, true);
+        styleJSON_.sprite = fixUrl(req, styleJSON_.sprite, item.publicUrl, false);
       }
       if (styleJSON_.glyphs) {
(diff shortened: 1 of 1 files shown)


petrsloup:
As stated in the comment two lines above -- the mapbox-gl-js cannot handle sprites with query.

This is because the sprite URL in the style looks like `http://..../sprite`, but the actual sprite images have several possible suffixes (`.png`, `.json`, `@2x.png`, `@2x.json`) and the javascript library simply concatenates these.

So if the style.json would contain `http://..../sprite?key=abc` the mapbox-gl-js would try to load `http://..../sprite?key=abc.png` (for example), which would fail.

candux:
This problem seems to be solved by this commit https://github.com/mapbox/mapbox-gl-js/commit/4f431b26ae0dab309c2868d930d51d651f07e10d from 2016 in mapbox-gl-js.

There is now a testcase for this sprite URLs like this: https://github.com/mapbox/mapbox-gl-js/blob/12991995fa1bb62c5ecdb0282c7c142578b897e5/test/unit/util/mapbox.test.js#L235

In case tileserver-gl wants to support clients before this change, then we could maybe introduce a config option for it.

petrsloup:
Oh, I guess I missed this. Merged.

PR #487, opened 2020-11-04
Render up to scale level 22

bheupers:
For detailed city rendering we need to render rasterfiles up to and including level 21. Therefore I propose to change the maximum raster level to 22


cvriel (approved):
(Helaas) noodzakelijk aanpassing in de standaard instellingen, maar goed gezien :-)

PR #488, opened 2020-11-04
update sharp to avoid segfault

vcabbage:
Before https://github.com/lovell/sharp/commit/e82a585cec48fa7a16e4edcccb1ef9021e27d2ca the server may crash with a segfault when under high load.

---

When testing with https://github.com/tsenart/vegeta found that the server would crash with a segfault in sharp/libvips. Narrowed the issue down to the commit referenced above.

Any version of sharp >=v0.25.2 resolves the problem.

Verified the issue and fix using the project's Dockerfile.

Command used for testing:
`echo 'GET http://localhost:8080/styles/basic-preview/14/8581/5738@2x.png' | vegeta attack -rate 50/s | vegeta report`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "pbf": "3.2.1",
     "proj4": "2.6.0",
     "request": "2.88.2",
-    "sharp": "0.25.1",
+    "sharp": "0.26.2",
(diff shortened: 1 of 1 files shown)


petrsloup:
Thanks!

PR #495, opened 2020-11-17
Add support for 512 sized raster tiles

Pesonet1:
This PR adds support for setting tile size as parameter for rendered tiles endpoint. Users should now be able to set tile size to be 256 or 512 pixels for rendered tiles when using direct endpoint for rendered tiles. Other functionality has been updated to use 256 pixel sized tiles as before changes.

Old rendered tiles endpoint:

`/styles/style/{z}/{x}/{y}[@2x].{format}`

New rendered tiles endpoint:

`/styles/style/{tileSize}/{z}/{x}/{y}[@2x].{format}`

Tile size can be 256 or 512 pixels.

WMTS getCapabilities document has been updated to contain two layers for serving 256 and 512 pixel sized tiles as separate layers. Additionally, tileMatrixSets have been added for 512 pixel sized tiles (EPSG:3857 and EPSG:4326).

**NOTE:** TileJSON file generation seems broken and these change will not show all possible endpoints in `tiles` section.

**NOTE:** Due to this change rendered tiles endpoint changes and it will break existing applications where direct rendered endpoint is used.

Ps. Feel free to open discussion to address some problems with it

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -13,10 +13,11 @@ Styles
 
 Rendered tiles
 ==============
-* Rendered tiles are served at ``/styles/{id}/{z}/{x}/{y}[@2x].{format}``
+* Rendered tiles are served at ``/styles/{id}/{tileSize}/{z}/{x}/{y}[@2x].{format}``
 
   * The optional ``@2x`` (or ``@3x``, ``@4x``) part can be used to render HiDPI (retina) tiles
   * Available formats: ``png``, ``jpg`` (``jpeg``), ``webp``
+  * Tile sizes: ``256``, ``512``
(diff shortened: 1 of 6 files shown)


EgorSaffrezzi:
Hi, when this will be implemented and merge ?

acalcutt:
I do not intend to merge this PR, it is too outdated and is incomplete.

I recently made https://github.com/maptiler/tileserver-gl/pull/1136 which is based off of this but on the current codebase. It adds some fixes to try and complete this. This one has a chance to be implemented, if the remaining issues can be solved there. I have no timeline, but any testers of that PR version would help speed it up.

PR #501, opened 2020-12-11
Fix URL to WMTS capabilities to be based on public URL

mloskot:
It's a trivial fix but very useful when the TileServer-GL is deployed behind a proxy and started with `--public_url`

### DIFF
--- a/public/templates/index.tmpl
+++ b/public/templates/index.tmpl
@@ -41,7 +41,7 @@
                 {{#if serving_data}}| {{/if}}<a href="{{public_url}}styles/{{@key}}.json{{&../key_query}}">TileJSON</a>
               {{/if}}
               {{#if serving_rendered}}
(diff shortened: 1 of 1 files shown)


hanneshdc:
I agree. We are deploying this server to aws, and have currently hit a roadblock because the `--public-url` setting does not affect the WMTS server.

mloskot:
It's been a while...Any chance to get it considered, reviewed and merged?

acalcutt (approved):
I think this change makes sense.

1.) If public url is not provided in options, it gets set to "/"
2.) All urls on this page use {{public_url}}, only this one does not.
3.) It has been requested in several PR, but this is the first.
4.) It has been used successfully in other forks (including my own).

mloskot:
Thank you @acalcutt

PR #506, opened 2020-12-18
Open mbtiles in readonly mode

xabbu42:
Open mbtiles in readonly mode as tileserver-gl never writes to them.
I'm not sure if there are any real benefits to this, but I figured it can't hurt.


acalcutt:
I think this one is worth adding. I used this in my fork and didn't have any issue with it. Are there any instances where tileserver would write back to the mbtiles file? I don't think I have seen any.

acalcutt (approved):
I think this is a good idea

xabbu42:
thanks, nice to see that tileserver-gl development is picking back up again.

PR #507, opened 2020-12-18
Use center and zoom from style json

xabbu42:
Use the center and zoom values from the style json as default for the center in the tilejson of the rendered map. This is especially useful for serveAllStyles where there is no other way to set this values explicitly.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -1375,7 +1375,7 @@ export const serve_rendered = {
     };
     const attributionOverride = params.tilejson && params.tilejson.attribution;
     if (styleJSON.center && styleJSON.zoom) {
-      tileJSON.center = styleJSON.center.concat(styleJSON.zoom);
+      tileJSON.center = styleJSON.center.concat(Math.round(styleJSON.zoom));
     }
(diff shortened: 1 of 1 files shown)


acalcutt (approved):

src/serve_rendered.js
```diff
@@ -1374,6 +1374,9 @@ export const serve_rendered = {
       type: 'baselayer',
     };
     const attributionOverride = params.tilejson && params.tilejson.attribution;
+    if (styleJSON.center && styleJSON.zoom) {
+      tileJSON.center = styleJSON.center.concat(styleJSON.zoom);
```
acalcutt, line 1378: I was going to merge this, but I had issues when the zoom level was a decimal value.
image

I figured out if I added a round Math.round aound styleJSON.zoom it fixes the issue
image

acalcutt: `tileJSON.center = styleJSON.center.concat(Math.round(styleJSON.zoom));` worked

acalcutt (changes requested):

acalcutt (approved):

PR #517, opened 2021-02-04
Fix docker run command in README

StefanBrand:
`tileserver-gl` actually listens on internal port `8080` (the port number on the right side of the colon).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Alternatively, you can use the `tileserver-gl-light` package instead, which is p
 An alternative to npm to start the packed software easier is to install [Docker](https://www.docker.com/) on your computer and then run in the directory with the downloaded MBTiles the command:
 
 ```bash
-docker run --rm -it -v $(pwd):/data -p 8080:80 maptiler/tileserver-gl
+docker run --rm -it -v $(pwd):/data -p 8080:8080 maptiler/tileserver-gl
 ```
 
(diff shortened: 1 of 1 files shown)


StefanBrand:
I see that in the `Dockerfile` port `80` is exposed:

https://github.com/maptiler/tileserver-gl/blob/c89a5ae029f5e9ef0855292cfcce2f72894a8fff/Dockerfile#L49

When I run `docker ps`, I get this under ports:

`80/tcp, 0.0.0.0:8080->8080/`

So, while in the browser I can access TileServer GL via `localhost:8080` when mapping `-p 8080:8080`, I can't do so when mapping `-p 8080:80`.

StefanBrand:
Maybe, this entry has to be edited as well:

https://github.com/maptiler/tileserver-gl/blob/c89a5ae029f5e9ef0855292cfcce2f72894a8fff/package.json#L17

lifeofguenter:
This might be a bug, the intention looks like it _should_ run on port 80: https://github.com/maptiler/tileserver-gl/blob/master/Dockerfile#L55

So ideally we update that as well, as there should be no reason on running it on a privileged port.

acalcutt:
This was fixed in https://github.com/maptiler/tileserver-gl/pull/602 . The docker now starts on port 80 inside like was intended and documented.

I'd submit another PR if there was interest in changing the ports.

PR #529, opened 2021-03-11
fix blocking of tileserver while serving index

candux:
we noticed sporadic long blocks of the entier tileserver on version 3.1.1
They always happend when somebody accessed the frontPage.
Our guess is, that clone() blocks the object and takes very long to finish.

We replaced the clone with ES6 magic from https://stackoverflow.com/a/42988763
and the problem is gone.

Also the time to generate the index goes from ~3s to ~0.2s
This should primarily affect user with many large styles

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -330,7 +330,7 @@ function start(opts) {
   };
 
   serveTemplate('/$', 'index', req => {
-    const styles = clone(serving.styles || {});
+    const styles = Object.assign({}, serving.styles || {}, {});
     for (const id of Object.keys(styles)) {
       const style = styles[id];
       style.name = (serving.styles[id] || serving.rendered[id] || {}).name;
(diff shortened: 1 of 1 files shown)


mnutt:
`Object.assign()` does a shallow copy, meaning only the top-level `styles` object is duplicated, while the `clone` module does a deep clone. Looking at the code following it, I think it's fine to leave the top-level alone and `Object.assign` the individual styles:

```javascript
serveTemplate('/$', 'index', req => {
    for (const id of Object.keys(serving.styles || {})) {
      const style = Object.assign({}, styles[id]);
      style.name = (serving.styles[id] || serving.rendered[id] || {}).name;
      style.serving_data = serving.styles[id];
      style.serving_rendered = serving.rendered[id];
      ...
```

Then the same can be done with `data`.

acalcutt:
This should be fixed by https://github.com/maptiler/tileserver-gl/pull/1009 . In that PR I have added a similar fix to index which does a more shallow array copy in https://github.com/WifiDB/tileserver-gl/blob/fa3dac46ca69ede31635867c4612dbf1a5b5decc/src/server.js#L450-L538

I am closing this PR now since it is no longer needed.

PR #544, opened 2021-05-21
Make subsequent calls of `npm run docker` faster

til-schneider:
This makes subsequent calls of `npm run docker` a lot faster if source code was changed.

How it works:
- The build step only executes `npm install`. So it only needs the `package.json` file.
- By copying the whole project folder as late as possible, docker can use cached layers for all previous steps. This includes long-running steps like `apt-get` and `npm install`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN export DEBIAN_FRONTEND=noninteractive \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


acalcutt:
I think the latest Dockerfile works pretty similar to this. Do you think this can be closed or is there something that we missed?

acalcutt:
I think our current docker file has most of what is here. please open a new PR if something is missing.

PR #555, opened 2021-10-25
Add .nvmrc file

olof-nord:
Due to this map tile server needing an old version of Node.js, I suggest to add a .nvmrc file.

This file can optionally be used by `nvm` (Node Version Manager) or `nvm-windows` to ease the node.js version used.

Once nvm is installed, this file can be used to install the specified Node.js version with `nvm install` and to activate Node.js with `nvm use`.

https://github.com/nvm-sh/nvm


olof-nord:
Thank you for the clarification @mnutt, I wasn't aware of the license switch.

PR #558, opened 2021-11-23
feat: allows resolving relative public urls into absolute urls

tVienonen:
Hello,

I had an issue with using tileserver-gl running behind a proxy in a subdirectory. I also wanted to configure my web clients with relative urls so that map server url was /maps. Mapbox GL requires that urls that come from tileserver-gl are absolute URLs.

This feature enables the server to serve content from subdirectory so that all the relative urls are resolved into absolute urls.

This is an opt-in feature so it won't be enabled by default. You have to specify TILESERVER_GL_RESOLVE_RELATIVE_PUBLIC_URL=true env variable and public_url option for it to be enabled.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -22,6 +22,12 @@ Getting started
     -s, --silent          Less verbose output
     -v, --version         Version info
 
+Resolve relative public url into an absolute url
+=====
+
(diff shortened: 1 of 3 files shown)


ochtec:
We have this same issue

vasily-polonsky:
any updates?

Issue #549, opened 2021-06-25
Bug with tileMargin setting at level 3


andreadapisa:
There is a bug when tileMargin is managed.
The first and last row of tiles are rendered wrong by mapbox-gl, because the requests is over the 85 degree (North or South).

In the follow image you can see a tile rendered with tileMargin = 80 (to avoid cropped texts) :
![image](https://user-images.githubusercontent.com/66669043/123455008-7ad3f080-d5e1-11eb-9c0a-339f3b2b0c14.png)

in the code I see tileMargin setting is used from level 3, but it should be avoid for the first and last row of tiles at every level.

The limitation at level 2 (tileMargin is not used) generates the cropped texts in the tiles:
![image](https://user-images.githubusercontent.com/66669043/123455918-8bd13180-d5e2-11eb-9cf5-a38384cd3007.png)

The tileMargin setting is useful, but the margin should be avoid to North for 0 row of tiles and to South for z²-1 row of tiles.

acalcutt:
Fixed by https://github.com/maptiler/tileserver-gl/pull/571

PR #571, opened 2022-02-22
Fix rendering with tileMargin on top and bottom border of map

frankstolle:
this fixes Bug with tileMargin setting at level 3 #549

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -302,9 +302,11 @@ module.exports = {
           });
 
           if (z > 2 && tileMargin > 0) {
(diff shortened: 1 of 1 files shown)


ungarj:
Thanks for the fix! We discovered this issue as well. Is there any chance to make a new release including this fix @petrsloup?

acalcutt (approved):
I tested this and confirmed the issue. without this fix in place and a TileMargin of 80, zoom level 3 looked like this

image

After applying this fix it looked like this
image

I tested various TileMargins (including 0) and different zoom levels and it didn't seem to cause any adverse affects.

PR #560, opened 2021-12-22
Simplify reloading the configuration with docker

zstadler:
Allow `docker kill -s HUP` and `docker-compose kill -s HUP` to reach the node process.

PR #575, opened 2022-03-05
Update usage.rst

zstadler:
Typo correction

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -35,4 +35,4 @@ Reloading the configuration
 It is possible to reload the configuration file without restarting the whole process by sending a SIGHUP signal to the node process.
 
 - The `docker kill -s HUP tileserver-gl` command can be used when running the tileserver-gl docker container.
-- The `docker-compose -s HUP tileserver-gl-service-name` can be used when tileserver-gl is run as a docker-compose service.
+- The `docker-compose kill -s HUP tileserver-gl-service-name` can be used when tileserver-gl is run as a docker-compose service.


zstadler:
Add a loop that keeps `docker-entrypoint.sh` runing after it passes a `HUP` signal to the tile server.

zstadler:
@petrsloup 
Could you merge this PR?
It fixes a bug in the original PR #560

pgassmann:
@petrsloup the latest image on dockerhub is from 2 years ago. Can you publish a updated image with these fixes?

docker-entrypoint.sh
```diff
@@ -24,6 +24,10 @@ if ! which -- "${1}"; then
 	# Wait exits immediately on signals which have traps set. Store return value and wait
 	# again for all jobs to actually complete before continuing.
 	wait $! || RETVAL=$?
+	while [ ${RETVAL} = 129 ] ; do
+	  # Refressh signal HUP received. Continue waiting for signals.
+	  wait $! || RETVAL=$?
+	done
```
acalcutt, line 30: @zstadler what does this actually do? why after HUP is received are we waiting for more signals?

acalcutt: but it doesn't actually reload the service, it just starts waiting again. I was thinking making a PR to replace it with something like this https://github.com/maptiler/tileserver-gl/issues/629#issuecomment-1292352424

zstadler:
```
trap refresh HUP
```
causes `refresh` to be executed when the shell process receives the HUP signal.

```
refresh() {
	SIGNAL=$(( $? - 128 ))
	echo "Caught signal ${SIGNAL}, refreshing"
	kill -s ${SIGNAL} $(pidof node) 2>/dev/null
}
```
causes the HUP signal to be sent to the `tileserver-gl` process.

```
wait $! || RETVAL=$?
```
keeps the shell process alive in order to relay any `HUP` signals to it.

```
while [ ${RETVAL} = 129 ] ; do
  # Refressh signal HUP received. Continue waiting for signals.
  wait $! || RETVAL=$?
done
```
The `wait` command exits once a signal is received. When a HUP signal is received, `RETVAL` is 129 and `wait` is executed again and the shell process waits for the next signal to come.

PR #580, opened 2022-03-28
CSS: Fix line break for long key-value strings

tordans:
`break-all` forces a line break for long strings (https://developer.mozilla.org/en-US/docs/Web/CSS/word-break).

before

after


mnutt (approved):

acalcutt:
this has to be updated a bit for the current version. mapbox-gl-inspect.css no longer exists, but the fix works in https://github.com/maptiler/tileserver-gl/blob/master/public/resources/maplibre-gl-inspect.css if 'mapbox-gl' is replaced with 'maplibregl'

acalcutt (approved):
I've used this and it works. I updated it for maplibre-gl-inspect

PR #583, opened 2022-04-04
fix problems with some urls when use --public_url

emiliopardo:
Hello

I've had some problems when trying to use tileserver-gl through a proxy.

These problems are related to the static files located in the path "/usr/src/app/public/"

The "wmts.tmpl" file does not contemplate the use of {{public_url}}, it always uses {{base_url}}

In the index.tmpl file in the wmts section, {{public_url}} is not considered in the link to the capabilities of the service.

In the index.css file all calls to styles that have url fail because they cannot be parameterized with {{public_url}}

If these styles are removed and included within index.tmpl in the header with <style> you can use {{public_url}}

These changes are minor and make it easier to use tileserver-gl through a proxy.

Thanks

### DIFF
--- a/public/templates/index.tmpl
+++ b/public/templates/index.tmpl
@@ -41,7 +41,7 @@
                 {{#if serving_data}}| {{/if}}<a href="{{public_url}}styles/{{@key}}.json{{&../key_query}}">TileJSON</a>
               {{/if}}
               {{#if serving_rendered}}
-                | <a href="/styles/{{@key}}/wmts.xml{{&../key_query}}">WMTS</a>
+                | <a href="{{public_url}}styles/{{@key}}/wmts.xml{{&../key_query}}">WMTS</a>
               {{/if}}
               {{#if xyz_link}}
                 | <a href="#" onclick="return toggle_xyz('xyz_style_{{@key}}');">XYZ</a>
--- a/public/templates/index.tmpl
+++ b/public/templates/index.tmpl
@@ -14,6 +14,37 @@
       return false;
     }
   </script>
+  <style>
+  @font-face {
+  font-family: "OpenSans";
+  src: url("{{public_url}}fonts/OpenSans-Regular.ttf");
+  font-weight: normal;
+  font-style: normal;
+}
+@font-face {
+  font-family: "OpenSans";
+  src: url("{{public_url}}fonts/OpenSans-Italic.ttf");
+  font-weight: normal;
+  font-style: italic;
+}
+@font-face {
+  font-family: "OpenSans";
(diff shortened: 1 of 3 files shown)


acalcutt:
I Merged https://github.com/maptiler/tileserver-gl/pull/605 , which sets BaseUrl to PublicUrl if the option is set. I think this should make the changes in wmts.tmpl not needed anymore.

I also merge https://github.com/maptiler/tileserver-gl/pull/501 , which fixes the public url in index

I think the only part left to this change would be the css, which I'm not sure about.

mnutt:
I believe the font-face definitions can be left in the css file, and use relative paths to the font files.

public/templates/index.tmpl
```diff
@@ -41,7 +72,7 @@
                 {{#if serving_data}}| {{/if}}<a href="{{public_url}}styles/{{@key}}.json{{&../key_query}}">TileJSON</a>
               {{/if}}
               {{#if serving_rendered}}
-                | <a href="/styles/{{@key}}/wmts.xml{{&../key_query}}">WMTS</a>
+                | <a href="{{public_url}}styles/{{@key}}/wmts.xml{{&../key_query}}">WMTS</a>
```
mloskot, line 75: I believe this has been fixed long ago
- #501

mloskot:
@tonofll 
> UP this PR Although API endpoints perfectly work by using `--public_url` parameter, certain UI elements and links are completely broken when you place `tileserver-gl` behind a proxy.

This is weird indeed, because the `public_url` seems to be used across the UI template, e.g. https://github.com/maptiler/tileserver-gl/blob/9bb270b6c577576324046c76303b0d20a9654326/public/templates/index.tmpl#L64-L72

There is a peculiar issue though:
- I'm using NGINX proxy
- I run pass `--public_url https://svc.example.com/test/tiles` (or `--public_url https://svc.example.com/test/tiles/`, no difference)
- I access `https://svc.example.com/test/tiles` and URL-s on the default frontpage are generated **without the part** `/tiles/` like `https://svc.example.com/test/styles/basic-preview/style.json` instead of
`https://svc.example.com/test/tiles/styles/basic-preview/style.json`
- I access `https://svc.example.com/test/tiles/` with trailing `/` and all URL-s on the default frontpage are correctly generated

The only issue I noticed is the missing thumbnails:
image

but that seems to be due to the URL-s missing `#`, I think, like here
`https://svc.example.com/test/tiles/styles/basic-preview/11/1072/717.png`
instead of
`https://svc.example.com/test/tiles/styles/basic-preview/#11/1072/717.png`

PR #593, opened 2022-07-15
591 - SQLite3 Vulnerability

### LINKED ISSUE #591: Update to use latest mbtiles for sqlite3 vulnerability
There is a NIST bulletin for a vulnerability in the nodejs-sqlite3 module that is used. https://nvd.nist.gov/vuln/detail/CVE-2022-21227

This is resolved in version >=5.0.2; however, the current release (3.1.1) of tileserver-gl uses `@mapbox/mbtiles` version `0.11.0` which is coded to 4.x module dependency for sqlite3. There was a later release of mbtiles almost 2 years ago even, 0.12.1 that utilizes `^5.0.0` which allows for easy update of the sqlite3 module to a fixed version.

ayohrling:
This updates the `@mapbox/mbtiles` version in package.json to utilize
`0.12.1` released about 2 years ago. This, in turn, utilizes `sqlite3`
version `5.x` which has patched vulnerabilities.

Closes #591

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@mapbox/glyph-pbf-composite": "0.0.3",
     "@mapbox/mapbox-gl-native": "5.0.2",
     "@mapbox/mapbox-gl-style-spec": "13.12.0",
-    "@mapbox/mbtiles": "0.11.0",
+    "@mapbox/mbtiles": "0.12.1",
     "@mapbox/sphericalmercator": "1.1.0",
     "@mapbox/vector-tile": "1.3.1",
     "advanced-pool": "0.3.3",


package.json
```diff
@@ -20,7 +20,7 @@
     "@mapbox/glyph-pbf-composite": "0.0.3",
     "@mapbox/mapbox-gl-native": "5.0.2",
     "@mapbox/mapbox-gl-style-spec": "13.12.0",
-    "@mapbox/mbtiles": "0.11.0",
+    "@mapbox/mbtiles": "0.12.1",
```
vinayakkulkarni, line 23: This causes breaking issue as v0.12.x has dropped support for Node 10.

https://github.com/mapbox/node-mbtiles/blob/master/CHANGELOG.md#0120

ayohrling: I mean, Node10 security support ended over a year ago, we really shouldn't be basing changes and updates (especially to vulnerability-related items) on past EOL Node versions...

ayohrling: This should be tested on 14,16,18 based on current support lifecycle.

acalcutt:
This should be fixed by https://github.com/maptiler/tileserver-gl/pull/602

As for the node 10 support, when I tested it it seemed to work, but with the move to maplibre-native we should probably suggest newer version of node anyway, like at least node 14 with 16 recommended.

petrsloup:
Fixed by #602

PR #596, opened 2022-08-08
style: fix lint issues in code 🕺

vinayakkulkarni:
- [x] Fixes issue with code styling in the repo. Linted the code via prettier. I plan to add `eslint` as well to identify [security](https://www.npmjs.com/package/eslint-plugin-security) vulnerabilities.. in coming PRs


vinayakkulkarni:
Thoughts on this @petrsloup ?

acalcutt:
I would maybe have this skip public/resources/ , since those are releases from outside this project

acalcutt:
I get the feeling looking at this that maybe it is overly complex. You mentioned a plan to use eslint eventually. couldn't that do most of this beautifying/formatting by itself? I feel like in my testing, eslint got the majority the formatting issues.

vinayakkulkarni:
> I get the feeling looking at this that maybe it is overly complex. You mentioned a plan to use eslint eventually. couldn't that do most of this beautifying/formatting by itself? I feel like in my testing, eslint got the majority the formatting issues.
> 
> Edit: For example, I feel like this was pretty close with just .eslintrc.yml and running eslint ** --fix
> 
> ```
> env:
>   browser: false
>   es2021: true
> extends: google
> overrides: []
> parserOptions:
>   ecmaVersion: latest
>   sourceType: module
> ignorePatterns:
>   - /public/resources
> rules:
>   comma-dangle:
>     - error
>     - never
> ```

https://github.com/vinayakkulkarni/tileserver-gl/blob/main/.eslintrc.cjs

added the newer version of `.eslintrc` since eslint themselves prefer using `.cjs` or `.js` file instead of `.yml`

also, why extend `google` when there are much better configs available. :( ?

PR #603, opened 2022-09-15
Update templates to use maplibre-gl-js, maplibre-gl-inspect, and leaflet . 

acalcutt:
This updates the templates to use maplibre-gl-js, maplibre-gl-inspect, and leaflet and gets rid of mapbox

There is a bit of hackiness in these templates to support older browsers, which basically loads IE compatible versions of the libraries if the browser doesn't support newer syntax. Without this, there is a browser supported check that is supposed to switch to a raster type that doesn't work
```
  <script>if (typeof Symbol !== 'undefined') { document.write('<script src="{{public_url}}maplibre-gl.js{{&key_query}}"><\/script>'); } else { document.write('<script src="{{public_url}}maplibre-gl-compat.js{{&key_query}}"><\/script>'); }</script>
  <script>if (typeof Symbol !== 'undefined') { document.write('<script src="{{public_url}}maplibre-gl-inspect.min.js{{&key_query}}"><\/script>'); } else { document.write('<script src="{{public_url}}maplibre-gl-inspect-compat.min.js{{&key_query}}"><\/script>'); }</script>
```


acalcutt:
If there are any recomendations please let me know. 

Right now, this is using the following libraries.

**maplibre-gl.js - Maplibre v2.4.0 (Official maplibre library)**
https://github.com/maplibre/maplibre-gl-js/releases

**maplibre-gl-compat.js - Maplibre v2.4.0 (old browser compatibility)**
https://github.com/acalcutt/maplibre-gl-js/tree/2.x_ie_compat

**maplibre-gl-inspect.min.js - mapbox-gl-inspect migrated to maplibre**
https://github.com/acalcutt/maplibre-gl-inspect

**maplibre-gl-inspect-compat.min.js - mapbox-gl-inspect migrated to maplibre (old browser compatibility)**
https://github.com/acalcutt/maplibre-gl-inspect/tree/internet-explorer

**leaflet.js - Leaflet 1.8.0 (offial leaflet library)**
https://leafletjs.com/download.html

**L.TileLayer.NoGap.js - Fix for leaflet tile gaps**
https://github.com/Leaflet/Leaflet.TileLayer.NoGap

petrsloup (approved):
Thank, nice update of the dependencies.

It would be nice to one day move past the committed built files and rather have maplibre-gl-js as an actual dependency and bundle during release. But that is definitely a bigger block of work for the future..

PR #605, opened 2022-09-15
Bugfix: publicUrl is not used in some urls

Monstrofil:
In WMTS the publicUrl path was missed and it was not working properly.

Similar story about wmts.tmpl: the baseUrl is not working properly when
tileserver is hidden behind the proxy with different root path.


acalcutt:
There are a few request related to this, like https://github.com/maptiler/tileserver-gl/pull/501 . Does anyone see any issues with this? it seems like it makes sense to me, but I'm not familiar enough with WMTS or the PublicURL option

acalcutt (approved):
I merged #501, so the PublicUrl portion of this has already been fixed. I think it makes sense to set baseUrl to be PublicUrl  when the user has that option set.  I also think changing the forward slashes to match the other templates makes sense.

PR #606, opened 2022-09-16
Convert to esm module santax

acalcutt:
Updates package to ESM module syntax, updates many packages . Replaced requires with imports.

One note relating to 'serve_light.js'. I had a lot of trouble making this so it would switch between full and light version without require. I ended up making this dummy file to load when 'serve_rendered.js' isn't needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "license": "BSD-2-Clause",
   "engines": {
-    "node": ">=14.13.0"
+    "node": ">=14.20.0 <17"
   },
   "scripts": {
     "test": "mocha test/**.js --timeout 10000",
@@ -25,7 +25,7 @@
     "@mapbox/sphericalmercator": "1.2.0",
     "@mapbox/vector-tile": "1.3.1",
     "advanced-pool": "0.3.3",
-    "canvas": "2.9.3",
+    "canvas": "2.10.1",
     "chokidar": "3.5.3",
     "clone": "2.1.2",
     "color": "4.2.3",
--- a/package.json
+++ b/package.json
@@ -4,13 +4,14 @@
   "description": "Map tile server for JSON GL styles - vector and server side generated raster tiles",
   "main": "src/main.js",
   "bin": "src/main.js",
+  "type": "module",
   "repository": {
     "type": "git",
-    "url": "https://github.com/maptiler/tileserver-gl.git"
+    "url": "git+https://github.com/maptiler/tileserver-gl.git"
   },
   "license": "BSD-2-Clause",
   "engines": {
-    "node": ">=10 <17"
+    "node": ">=14.13.0"
   },
   "scripts": {
     "test": "mocha test/**.js --timeout 10000",
@@ -25,25 +26,24 @@
     "@mapbox/vector-tile": "1.3.1",
     "advanced-pool": "0.3.3",
     "canvas": "2.9.3",
-    "chokidar": "3.3.1",
+    "chokidar": "3.5.3",
     "clone": "2.1.2",
-    "color": "3.1.2",
-    "commander": "4.1.1",
+    "color": "4.2.3",
+    "commander": "9.4.0",
     "cors": "2.8.5",
-    "esm": "3.2.25",
-    "express": "4.17.1",
-    "handlebars": "4.7.3",
+    "express": "4.18.1",
+    "handlebars": "4.7.7",
     "http-shutdown": "1.2.2",
-    "morgan": "1.9.1",
+    "morgan": "1.10.0",
     "pbf": "3.2.1",
-    "proj4": "2.6.0",
+    "proj4": "2.8.0",
     "request": "2.88.2",
-    "sharp": "0.26.2",
+    "sharp": "0.30.7",
     "tileserver-gl-styles": "2.0.0"
   },
   "devDependencies": {
-    "mocha": "^7.1.0",
-    "should": "^13.2.3",
-    "supertest": "^4.0.2"
+    "mocha": "^10.0.0",
+    "chai": "4.3.6",
+    "supertest": "^6.2.4"
   }
 }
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -24,26 +25,25 @@
     "@mapbox/sphericalmercator": "1.2.0",
     "@mapbox/vector-tile": "1.3.1",
     "advanced-pool": "0.3.3",
-    "canvas": "2.9.3",
```
vinayakkulkarni, line 27: For dependencies management, we should use dependabot or renovate bot instead of manually updating the deps

src/main.js
```diff
@@ -2,22 +2,25 @@
 
 'use strict';
 
-require = require('esm')(module);
+import fs from 'fs';
```
vinayakkulkarni, line 5:
```suggestion
import fs from 'node:fs';
```


Maybe we can add (idk what this is called, namespace?) namespaces?

src/main.js
```diff
@@ -118,49 +121,48 @@ const startWithMBTiles = (mbtilesFile) => {
       }
       const bounds = info.bounds;
 
-      const styleDir = path.resolve(__dirname, "../node_modules/tileserver-gl-styles/");
+      const styleDir = path.resolve(__dirname, '../node_modules/tileserver-gl-styles/');
 
       const config = {
-        "options": {
```
vinayakkulkarni, line 124: We should use standard Eslint & Prettier configs for auto cleanup of styling...

package.json
acalcutt: I agree. before there were a few dependencies that were stuck at old versions due to the old version of node, but if we raise the minimum node version that becomes a bit easier and we can automatically update a few things

src/main.js
acalcutt: I ran eslint on this, but could use suggestions on defaults

src/main.js
acalcutt: Is that preferred? I'm not to familiar with that, but it would be a simple change. i think that exists in a few files.

package.json
acalcutt: I removed unnecessary package updates from this pr. we can update them later with a more official method. there were a few devdeps relating to test i left and commander which I updated the syntax.

mnutt:
It seems like given static es6 imports, maybe the 'right' way to do the light version would be to have a separate entrypoint into the app. It'd probably be a lot to add now, but maybe in the future with some refactoring.

acalcutt:
I'm definitely up for suggestions on the light version. I had tried a few ideas I had found online, but this is the one I ended up getting to work. I was just happy to not have to get rid of that feature at the time. 

To be honest I'm more a general purpose scripting programmer than a javascript programmer, so I am always open to tips of how i can do things better.

src/main.js
mnutt: Same as below, these keys could actually have the quotes removed entirely since they're js rather than JSON.

mnutt: Also, I'm a fan of eslint/prettier; it ends up touching most lines of the project so I'd suggest getting anything in-progress merged before running it to reduce the amount of merge conflicts.

src/serve_data.js
```diff
@@ -1,18 +1,18 @@
...
+import express from 'express';
+import MBTiles from '@mapbox/mbtiles';
+import Pbf from 'pbf';
+import VectorTile from '@mapbox/vector-tile';
 
-const utils = require('./utils');
+import {getTileUrls, fixTileJSONCenter} from './utils.js';
 
-module.exports = {
+export const serve_data = {
```
mnutt, line 15: Maybe also not for this PR, but we might consider replacing all of the snake_case exports with camelCase ones.

src/server.js
```diff
@@ -1,52 +1,53 @@
...
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const packageJson = JSON.parse(fs.readFileSync(__dirname + '/../package.json', 'utf8'));
+
 const isLight = packageJson.name.slice(-6) === '-light';
-if (!isLight) {
-  // do not require `serve_rendered` in the light package
-  serve_rendered = require('./serve_rendered');
-}
+const serve_rendered = (await import(`${!isLight ? `./serve_rendered.js` : `./serve_light.js`}`)).serve_rendered;
```
mnutt, line 31: I think this is actually the optimal solution for now.

The other idea I floated was to have two entrypoints (`src/main.js` and `src/light.js` or something) and `main.js` would import `server_rendered.js` while `light.js` would include the stub, then they each would import `src/server.js` and pass `serve_rendered` method to it. But in its current state it'd be a lot of duplication and a fair amount of passing `serve_rendered` around, or a global variable or something. Neither is better than what we have here, IMO.

mnutt (changes requested):

src/serve_rendered.js
```diff
@@ -574,33 +570,33 @@ module.exports = {
           ratio: ratio,
```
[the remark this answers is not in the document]
acalcutt: updated it to mlgl

src/main.js
```diff
@@ -118,49 +121,48 @@ const startWithMBTiles = (mbtilesFile) => {
       }
       const bounds = info.bounds;
 
-      const styleDir = path.resolve(__dirname, "../node_modules/tileserver-gl-styles/");
+      const styleDir = path.resolve(__dirname, '../node_modules/tileserver-gl-styles/');
```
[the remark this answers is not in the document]
acalcutt: is this require.resolve depreciated? it seems like it is gone in node 18  https://nodejs.org/api/esm.html#no-requireresolve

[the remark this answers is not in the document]
mnutt: Ah right. And `import.meta.resolve()` is behind an experimental flag even in latest node 18, so probably a nonstarter.

mnutt (approved):

vinayakkulkarni:
This is awesome !

Thanks for the updates :)

PR #608, opened 2022-09-21
Add static renderer pool to fix static images

acalcutt:
This adds a second pool of renderers that are in static mode instead tile mode. This works around a change in maplibre-native that only returns one tile when in tile mode, which causes static images with multiple tiles to fail to load completely.

Without this change, tileserver reports the following warning
```
mbgl: {
  class: 'General',
  severity: 'WARNING',
  text: 'Provided camera options returned 4 tiles, only 11/1100/671=>11 is taken in Tile mode.'
}
```
and the image looks like this
500x500_old

After this change, the warning does not show and the static image looks like this
500x500_new

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -229,9 +229,7 @@ module.exports = {
 
     const app = express().disable('x-powered-by');
 
-    const respondImage = (item, z, lon, lat, bearing, pitch,
-      width, height, scale, format, res, next,
-      opt_overlay) => {
+    const respondImage = (item, z, lon, lat, bearing, pitch, width, height, scale, format, res, next, opt_overlay, opt_mode="tile") => {
       if (Math.abs(lon) > 180 || Math.abs(lat) > 85.06 ||
         lon !== lon || lat !== lat) {
         return res.status(400).send('Invalid center');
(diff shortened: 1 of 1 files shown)


acalcutt:
Anyone have any input on this? I may abandon the attempt to revert https://github.com/maplibre/maplibre-gl-native/commit/9ceb619940901d06909f7d6d0fc4a4e64572d228 on maplibre-native in favor of changing it here. 

I'd like to get this static image issue fixed before we update the docker image, so any reviews of this before I merge it would be nice.

acalcutt:
I've discussed this a bit with Birk Skyum on the maplibre side and we have decided maplibre returning a single tile when in tile mode is expected behavior and not a bug. 

This change works around that limitation by creating a 'static' mode rendering pool for use by the static endpoints. tile mode is still used when tileserver-gl is working as a tile server, which is needed to prevent label clipping.

I'm going to merge this.

nyurik:
Hi @acalcutt, I absolutely love that you are working on this project (thanks!!!), but for the overall health of it it would be awesome if someone (even if they are not on the project) would review it to spot some silly things and to do a knowledge transfer. Otherwise it will basically be just you who knows everything about it.  Can you think of anyone else who might be able to do it, and maybe add them as a reviewer to your PRs?

vinayakkulkarni:
I am still facing this issue with latest v4.2.0 release.

PR #613, opened 2022-09-24
Don't include duplicate attributions

mnutt:
Check to see if tileJSON already has a particular attribution before appending it again.

Longer term I'd probably keep `attributions` in an array and `join(' | ')` at the end, but we'd probably need some more cleanup before that is viable.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -792,10 +792,12 @@ module.exports = {
 
               if (!attributionOverride &&
                 source.attribution && source.attribution.length > 0) {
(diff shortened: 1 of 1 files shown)


acalcutt (approved):
Fixes duplicate attributions which looks like this 

duplicate_attributes

acalcutt:
Just an FYI on this

The behavior was made to be similar to maplibre attributions. The maplibre code uses and array and joing like you mention. ( https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/control/attribution_control.ts#L158 )

PR #619, opened 2022-09-29
Extended Static-Images Endpoint

benedikt-brandtner-bikemap:
I have Extended the Static-Images Endpoint with the following capabilities:

- Optional border with configurable color and width for paths via the `border` and `borderwidth` parameters
- Optional markers with the following configurable options:
  - Icon provided locally via the icons path or as a hyperlink (enabled via `allowRemoteMarkerIcons` setting)
  - Scale of the image relative to its original size
  - Offset in pixels
- Linecap of the path https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap
- Linejoin of the path https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin
- Capability to provide multiple paths
- Maximum zoom level if autofit is selected when providing a path

Additional Changes:

- In order to keep performance drawing of elements has been refactored to be asynchronous
- Refactored a lot of helpers in serve_rendered.js to be more modular and reusable

Please let me know if this would be interesting to you.

Cheers :)

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -38,6 +38,7 @@ Static images
   * ``path`` - comma-separated ``lng,lat``, pipe-separated pairs
 
     * e.g. ``5.9,45.8|5.9,47.8|10.5,47.8|10.5,45.8|5.9,45.8``
+    * can be provided multiple times
 
   * ``latlng`` - indicates coordinates are in ``lat,lng`` order rather than the usual ``lng,lat``
   * ``fill`` - color to use as the fill (e.g. ``red``, ``rgba(255,255,255,0.5)``, ``#0000ff``)
@@ -63,8 +64,8 @@ Static images
         * scales with ``scale`` parameter since image placement is relative to it's size
         * e.g. ``2,-4`` - Image will be moved 2 pixel to the right and 4 pixel in the upwards direction from the provided location
 
-    * can be provided multiple times
     * e.g. ``5.9,45.8|marker-start.svg|scale:0.5|offset:2,-4``
+    * can be provided multiple times
 
   * ``padding`` - "percentage" padding for fitted endpoints (area-based and path autofit)
 
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -39,13 +39,37 @@ Static images
 
     * e.g. ``5.9,45.8|5.9,47.8|10.5,47.8|10.5,45.8|5.9,45.8``
 
-  * ``latlng`` - indicates the ``path`` coordinates are in ``lat,lng`` order rather than the usual ``lng,lat``
+  * ``latlng`` - indicates coordinates are in ``lat,lng`` order rather than the usual ``lng,lat``
   * ``fill`` - color to use as the fill (e.g. ``red``, ``rgba(255,255,255,0.5)``, ``#0000ff``)
   * ``stroke`` - color of the path stroke
   * ``width`` - width of the stroke
+  * ``linecap`` - line cap of the path stroke
+  * ``border`` - color of the optional border path stroke
+  * ``borderwidth`` - width of the border stroke (default 10% of width)
+  * ``marker`` - Marker in format ``lng,lat|iconPath|option|option|...``
+
+    * Will be rendered with the bottom center at the provided location
+    * ``lng,lat`` and ``iconPath`` are mandatory and icons won't be rendered without them
+    * ``iconPath`` is either a link to an image served via http(s) or a path to a file relative to the configured icon path
+    * ``option`` must adhere to the format ``optionName:optionValue`` and supports the following names
+
+      * ``scale`` - Factor to scale image by
+
+        * e.g. ``0.5`` - Scales the image to half it's original size
+
+      * ``offset`` - Image offset as positive or negative pixel value in format ``[offsetX],[offsetY]``
+
+        * scales with ``scale`` parameter since image placement is relative to it's size
+        * e.g. ``2,-4`` - Image will be moved 2 pixel to the right and 4 pixel in the upwards direction from the provided location
+
+    * can be provided multiple times
+    * e.g. ``5.9,45.8|marker-start.svg|scale:0.5|offset:2,-4``
+
   * ``padding`` - "percentage" padding for fitted endpoints (area-based and path autofit)
 
     * value of ``0.1`` means "add 10% size to each side to make sure the area of interest is nicely visible"
+
+  * ``maxzoom`` - Maximum zoom level (only for auto endpoint where zoom level is calculated and not provided)
 
 * You can also use (experimental) ``/styles/{id}/static/raw/...`` endpoints with raw spherical mercator coordinates (EPSG:3857) instead of WGS84.
 
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -43,7 +43,8 @@ Static images
   * ``fill`` - color to use as the fill (e.g. ``red``, ``rgba(255,255,255,0.5)``, ``#0000ff``)
   * ``stroke`` - color of the path stroke
   * ``width`` - width of the stroke
-  * ``linecap`` - line cap of the path stroke
+  * ``linecap`` - rendering style for the start and end points of the path
+  * ``linejoin`` - rendering style for overlapping segments of the path with differing directions
   * ``border`` - color of the optional border path stroke
   * ``borderwidth`` - width of the border stroke (default 10% of width)
   * ``marker`` - Marker in format ``lng,lat|iconPath|option|option|...``
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -8,10 +8,10 @@ import util from 'util';
 import zlib from 'zlib';
 import sharp from 'sharp'; // sharp has to be required before node-canvas. see https://github.com/lovell/sharp/issues/371
 import pkg from 'canvas';
-import Image from 'canvas';
 import clone from 'clone';
 import Color from 'color';
 import express from 'express';
+import sanitize from "sanitize-filename";
 import SphericalMercator from '@mapbox/sphericalmercator';
 import mlgl from '@maplibre/maplibre-gl-native';
 import MBTiles from '@mapbox/mbtiles';
@@ -22,7 +22,7 @@ import {getFontsPbf, getTileUrls, fixTileJSONCenter} from './utils.js';
 const FLOAT_PATTERN = '[+-]?(?:\\d+|\\d+\.?\\d+)';
 const httpTester = /^(http(s)?:)?\/\//;
 
-const {createCanvas} = pkg;
+const {createCanvas, Image} = pkg;
 const mercator = new SphericalMercator();
 const getScale = (scale) => (scale || '@1x').slice(1, 2) | 0;
 
@@ -231,11 +231,20 @@ const extractMarkersFromQuery = (query, options, transformer) => {
     // Check if icon is served via http otherwise marker icons are expected to
     // be provided as filepaths relative to configured icon path
     if (!(iconURI.startsWith('http://') || iconURI.startsWith('https://'))) {
-      iconURI = path.resolve(options.paths.icons, iconURI);
-      // Ensure icon exists at provided path
-      if (!fs.existsSync(iconURI)) {
+      // Sanitize URI with sanitize-filename
+      // https://www.npmjs.com/package/sanitize-filename#details
+      iconURI = sanitize(iconURI)
+
+      // If the selected icon is not part of available icons skip it
+      if (!options.paths.availableIcons.includes(iconURI)) {
         continue;
       }
+
+      iconURI = path.resolve(options.paths.icons, iconURI);
+
+    // When we encounter a remote icon check if the configuration explicitly allows them.
+    } else if (options.allowRemoteMarkerIcons !== true) {
+      continue;
     }
 
     // Ensure marker location could be parsed
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -135,23 +135,57 @@ const parseCoordinates = (coordinatePair, query, transformer) => {
   return parsedCoordinates;
 };
 
-const extractPathFromQuery = (query, transformer) => {
-  const pathParts = (query.path || '').split('|');
-  const path = [];
-  for (const pair of pathParts) {
-    const pairParts = pair.split(',');
-    if (pairParts.length === 2) {
-      const pair = parseCoordinates(pairParts, query, transformer);
 
-      // Ensure coordinates could be parsed and skip them if not
-      if (pair === null) {
-        continue;
-      }
-
-      path.push(pair);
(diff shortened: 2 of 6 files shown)


Dockerfile_test
```diff
@@ -4,22 +4,35 @@
...
+      libgbm-dev \
+      libprotobuf-dev \
+      libegl1 \
+      xvfb \
+      xauth \
+      libopengl0 \
+      libcurl4 \
+      libuv1-dev \
+      libc6-dev \
+      libcap2-bin \
```
acalcutt, line 27: Does the docker need these new dependencies?

Dockerfile_test
```diff
@@ -4,22 +4,35 @@
...
+  && apt-get -y --purge autoremove \
+  && apt-get clean \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN curl http://archive.ubuntu.com/ubuntu/pool/main/libj/libjpeg-turbo/libjpeg-turbo8_2.0.3-0ubuntu1_amd64.deb --output libjpeg-turbo8_2.0.3-0ubuntu1_amd64.deb
```
mnutt, line 32: Is there a reason we need to download this manually vs apt-getting it? Is it a versioning thing?

docs/endpoints.rst
```diff
@@ -39,14 +39,38 @@ Static images
...
+    * Will be rendered with the bottom center at the provided location
+    * ``lng,lat`` and ``iconPath`` are mandatory and icons won't be rendered without them
+    * ``iconPath`` is either a link to an image served via http(s) or a path to a file relative to the configured icon path
+    * ``option`` must adhere to the format ``optionName:optionValue`` and supports the following names
+
+      * ``scale`` - Factor to scale image by
+
+        * e.g. ``0.5`` - Scales the image to half it's original size
+
+      * ``offset`` - Image offset as positive or negative pixel value in format ``[offsetX],[offsetY]``
```
mnutt, line 60: I'm not opposed to this by any means, but have you considered just requiring that all markers be centered both horizontally and vertically on the coordinates? IMO it makes it really easy to get placement right when designing icons, and works really well with scaling, etc.

src/serve_rendered.js
```diff
@@ -93,37 +94,337 @@ function createEmptyResponse(format, color, callback) {
...
+    // Ensure the locationParts contains two items
+    if (locationParts.length !== 2) {
+      continue;
+    }
+
+    let iconURI = markerParts[1];
+    // Check if icon is served via http otherwise marker icons are expected to
+    // be provided as filepaths relative to configured icon path
+    if (!(iconURI.startsWith('http://') || iconURI.startsWith('https://'))) {
+      iconURI = path.resolve(options.paths.icons, iconURI);
```
mnutt, line 234: I believe this would let someone specify a relative path that broke out of the icons dir and could pull from anywhere in the system. Maybe extract a function that checks that `iconURI.startsWith(options.paths.icons)`?

src/serve_rendered.js
```diff
@@ -93,37 +94,337 @@ function createEmptyResponse(format, color, callback) {
...
+      // Ensure icon exists at provided path
+      if (!fs.existsSync(iconURI)) {
```
mnutt, line 236: We likely want to see if we can somehow avoid `existsSync` since it's in the request path. One way would be on startup to just synchronously pull a list of images in the icons dir, put them in an object, and retrieve their full paths here. It'd also solve the above issue around limiting the paths to the icons dir.

mnutt (commented):
Thank you for this contribution! I'll be looking through it more thoroughly but wanted to leave a few initial comments.

src/serve_rendered.js
boldtrn: I think we should definitely try to sanitize the input here. I would like to propose something like this: https://github.com/acalcutt/tileserver-gl/commit/64cad7da66eb1af01d079c547d69b2af990357d8

Not sure if we might need to be stricter and catch even more?

boldtrn (changes requested):
Thank you very much for this great PR @benedikt-brandtner-bikemap. The features are really awesome 👍 .

I personally think that we should add a bit of security logic to this PR. Maybe we could make requesting external images configureable. For example if this is a public server, a malicious user might host a `https://` image somewhere and get the server to download this. At the very least it might be possible to force the server to download huge files and essentially DDOS the server. 

The same applies to the file name, in theory a malicious user could request some sensitive files and draw them on the map or in the worst case might even run some code injection.

WDYT?

mnutt:
There's also the tricky concept of SSRF--allowing a server to take instruction to download any URL means it can fetch _internal_ resources that the server has access to (ec2 instances often have access to sensitive metadata on localhost, etc) and it's not very easy to set up rules to prevent it.

Realistically, siphoning internal resources through embedded markers is probably pretty difficult but would still potentially allow someone to prod around. I think it'd probably be best to leave the remote feature disabled by default, and let people enable it knowing the security implications.

acalcutt:
Should this use a similar syntax to maptilers other products, like https://documentation.maptiler.com/hc/en-us/articles/360020805897-Static-maps-for-your-web . Seems like they support icon urls

Dockerfile_test
benedikt-brandtner-bikemap: It seems like mablibre-gl depends on a version of libjpeg-turbo with legacy-capabilities enabled for version 8 which doesn't seem to be available in the current apt repository and also a libicu version which is not in the repository.

But for the dockerfile i just copied the packages from `Dockerfile`

docs/endpoints.rst
benedikt-brandtner-bikemap: Maybe I don't understand the question, but the markers are already centered above the selected location https://github.com/maptiler/tileserver-gl/pull/619/files#diff-e18e0ca2c1587af2fb3550206da0889f4297c04bf4e3a1d1ee09ca46041a3c3dR51 and https://github.com/maptiler/tileserver-gl/pull/619/files#diff-51ea374a3960babfb87e15ca326fc54b011b08f981c33da61d174022970a5ce6R286
I guess it would be nice to have a `anchor` parameter similar to https://documentation.maptiler.com/hc/en-us/articles/360020805897-Static-maps-for-your-web#Staticmapsforyourweb-Addingmarkers but idk if that should be in the scope of this PR

acalcutt:
Hope you are feeling better @benedikt-brandtner-bikemap . No rush here.

src/serve_rendered.js
benedikt-brandtner-bikemap: have added the path sanitation with `sanitize-filename`

src/serve_rendered.js
benedikt-brandtner-bikemap: all icons are now loaded into a settings object on server startup and the provided icon is checked against this list

benedikt-brandtner-bikemap:
thanks a lot, but yeah am bored anyways so this is a welcome distraction :slightly_smiling_face: 

Have implemented the following fixes to address the mentioned issues:

1. Loading of remote images must now be allowed via the configuration parameter `allowRemoteMarkerIcons`
2. Locally available icons get loaded as relative paths to the icon-folder into the settings object `paths.availableIcons` on server startup (thanks a lot for the idea @mnutt)
3. When providing a local icon as a marker, the path gets sanitized first and then checked against the `paths.availableIcons` settings object if the icon exists and is allowed to be used

I think these measures should remedy a lot of the security concerns mentioned, although you are correct as soon as the icons get fetched from a remote server we are still exposed.

benedikt-brandtner-bikemap:
Have rebased my changes ontop of current master to resolve conflicts in `Dockerfile_test`

src/server.js
```diff
@@ -92,6 +93,36 @@ export function server(opts) {
...
+      const entryPath = path.resolve(directory, dirEntry.name);
+      return dirEntry.isDirectory() ?
+        getFiles(entryPath) : entryPath.replace(paths.icons + path.sep, "");
+    }));
+
+    // Flatten the list of files to a single array
+    return files.flat();
+  }
+
+  // Load all available icons into a settings object
```
boldtrn, line 119: Great idea 👍

boldtrn:
> I think these measures should remedy a lot of the security concerns mentioned, although you are correct as soon as the icons get fetched from a remote server we are still exposed.

Thanks a lot for the changes. At least from my side, this solves the concerns I had and I am looking forward to using this 👍

acalcutt:
I finally got a chance to set up a test of this. Seemed to work pretty well. it took me a bit to figure out the url.

I ended up with this for my test url
http://192.168.0.251:8080/styles/test-style/static/auto/800x600.png?&marker=5.9,45.9|icons8-maps.svg&marker=5.7,44.9|icons8-pin-62.png|scale:0.75&marker=5,45.1|icons8-pin-67.png|scale:0.75
and got
800x600

acalcutt:
How do you use linecap?

benedikt-brandtner-bikemap:
Hey, awesome that you approve of my changes if i should add/change anything else i'll be happy to do so!

@acalcutt I use linecap in the drawing context of the path if one is provided to define the rendering-style of the start- and end-points of the line as described [here](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap)
[implementation](https://github.com/maptiler/tileserver-gl/pull/619/files#diff-51ea374a3960babfb87e15ca326fc54b011b08f981c33da61d174022970a5ce6R414) 

Also you can find documentation for the new staticmap endpoint parameters (including markers) [here](https://github.com/maptiler/tileserver-gl/pull/619/files#diff-e18e0ca2c1587af2fb3550206da0889f4297c04bf4e3a1d1ee09ca46041a3c3d)

benedikt-brandtner-bikemap:
Have updated documentation for the linecap parameter and also added a linejoin parameter as well to be able to alter the rendering style of overlapping segments as described [here](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin)

acalcutt:
Based on the docs and links, shouldn't this give me squared off line endings?

http://192.168.0.251:8080/styles/test-style/static/auto/800x600.png?&path=5.9,45.3|5.7,45.2&linecap=square
800x600 (3)

benedikt-brandtner-bikemap:
Seems to work maybe you just cant see it because the width is too thin

default (butt):
http://127.0.0.1:8080/styles/bikemap-2021-3D/static/auto/800x600.png?width=10&path=16.370079517364502,48.2174529909167|16.3616681098938,48.21362115453811
default

square:
http://127.0.0.1:8080/styles/bikemap-2021-3D/static/auto/800x600.png?width=10&linecap=squ
are&path=16.370079517364502,48.2174529909167|16.3616681098938,48.21362115453811
square

round:
http://127.0.0.1:8080/styles/bikemap-2021-3D/static/auto/800x600.png?width=10&linecap=round&path=16.370079517364502,48.2174529909167|16.3616681098938,48.21362115453811
round


Also i've added the capability to draw multiple paths, which i wanted to do when i initially wrote the extensions but which came up again now.

sorry for extending the scope of this PR but it was only a minor change on top

acalcutt:
Your right, it does work. i guess I expected more of a square at the ends. looks good.

acalcutt:
@mnutt and @boldtrn  Any other things you would like to see before this is merged. it seems like most of the complaints have been fixed and it seems to work pretty well. I'd approve it and merge it in.

the only thing I wish we had was a default marker icon in our test style...but I think that would be something I would have to add to the old release zip file we use.

boldtrn:
Default Marker would be nice indeed. So far I think this PR seems nice and I'd love to start using it.

acalcutt:
I've been thinking a bit about the use case where you give tileserver a mbtiles file but do not give it a config. In that case it uses the built in style in '/node_modules/tileserver-gl-styles/', so it makes it somewhat hard to add the marker icons unless you add a config.

I was thinking, what if we put some default marker images in 'public/resources/icons/' and in main.js , add that icon path to the default generated config like this around line 125
```
      const styleDir = path.resolve(__dirname, '../node_modules/tileserver-gl-styles/');
      **const iconDir = path.resolve(__dirname, 'public/resources/icons/');**

      const config = {
        'options': {
          'paths': {
            'root': styleDir,
            'fonts': 'fonts',
            **'icons': iconDir,**
            'styles': 'styles',
            'mbtiles': path.dirname(mbtilesFile)
          }
        },
        'styles': {},
        'data': {}
      };
```

acalcutt:
maybe something CC0 licensed for a default icon, like https://uxwing.com/map-pin-icon/

acalcutt:
I was also thinking I wonder if the icons should be more like sprites so you could use a single image like https://github.com/sigma-geosistemas/Leaflet.awesome-markers/blob/master/dist/images/markers-matte.png with multiple colors in one file.

Not sure its needed to merge this, I was just wondering if that would be better. it seems like sprites are very similar to markers.

mnutt:
I've looked over it and am good with it.

Regarding default markers: for what it's worth my old branch has a default marker that can be colorized based on a hex code:

https://github.com/mnutt/tileserver-gl/blob/master/src/markers/color.js

The 'template' uses solid red as the replacement color, and will dynamically generate PNGs of other colors.

acalcutt:
> I've looked over it and am good with it.
> 
> Regarding default markers: for what it's worth my old branch has a default marker that can be colorized based on a hex code:
> 
> https://github.com/mnutt/tileserver-gl/blob/master/src/markers/color.js
> 
> The 'template' uses solid red as the replacement color, and will dynamically generate PNGs of other colors.

That is a really nice feature.

benedikt-brandtner-bikemap:
Sorry for not beeing more active was focused on other tasks..

Is there anything I can do?

greets

acalcutt:
What did you think of https://github.com/maptiler/tileserver-gl/pull/619#issuecomment-1279773329 , adding a default icon path.

acalcutt (approved):
Looks good to me. I still would like to see a default marker included, just so this can be used out of the box with the built in styles (like when just loading a mbtiles file with no config). then we could also use that default marker name in the documentation.

benedikt-brandtner-bikemap:
Hey, sorry for not answering sooner was again occupied with something else...

Hmm I could implement a default marker with a choosable color utilizing @mnutt lib but I think I'll only have time end of next week so maybe we merge this PR and I'll open a new one with a default marker?

Greets

acalcutt:
that sounds fine to me. I'll merge this.

acalcutt:
Hey @benedikt-brandtner-bikemap ,

I noticed this after I merged, but would it be possible to make a PR that changes 

`import pkg from 'canvas'; `https://github.com/maptiler/tileserver-gl/blob/master/src/serve_rendered.js#L10
to
`import {createCanvas, Image} from 'canvas';`

and remove the line
`const {createCanvas, Image} = pkg;` https://github.com/maptiler/tileserver-gl/blob/master/src/serve_rendered.js#L25

benedikt-brandtner-bikemap:
sorry just saw your comment, thanks for fixing this and merging :)!

PR #628, opened 2022-10-26
add arm64 docker build to workflow (once maplibre binary is available)

acalcutt:
Creates a workflow to test and publish to npm and dockerhub

Once there is a arm64 compatible maplibre, this build workflow will make a multi-platform tileserver-gl docker release. It is set up to emulate arm64 using QEMU so the docker image can be built.

Right now I don't have the ability to add environment secrets, so I have added text inputs for Docker Username/Token and NPM Token  needed to publish.


.github/workflows/Build_Test_Release.yml
```diff
@@ -0,0 +1,109 @@
...
+    inputs:
+      docker_user:
+        description: 'Docker Username'     
+        required: true
+      docker_token:
+        description: 'Docker Token'     
+        required: true
+      npm_token:
+        description: 'NPM Token'
+        required: true
```
vinayakkulkarni, line 14: I clean up and set this up in the #597 ?

vinayakkulkarni: - [x] Done in e47cb1d

mnutt (approved):

Issue #629, opened 2022-10-26
Using SIGHUP to restart is not working


trondat:
Hi,

I have used SIGHUP to restart the server earlier, this has worked fine. With the current verison it just responds with "Hangup" and then the server stops responding. 

I'm running tileserver-gl in Docker and sends the command to the server like this :
`docker exec tileserver-gl bash -c 'kill -HUP $(ls -l /proc/*/exe | sed -n "/\/node$/s/.*proc\/\([0-9]\+\)\/exe .*/\1/p")' `

This gives : 
```Starting server
Listening at http://[::]:80/
Startup complete
Hangup
```
After this it stops responding.

Is there some way of working around this ?

acalcutt:
This seems to be related to https://github.com/maptiler/tileserver-gl/pull/575

If I remove this section added to docker-entrypoint.sh, the process ends.
```
	while [ ${RETVAL} = 129 ] ; do
		echo 
	  # Refressh signal HUP received. Continue waiting for signals.
	  wait $! || RETVAL=$?
	done
```

acalcutt:
the is some updated documentation relating to that at https://tileserver.readthedocs.io/en/latest/usage.html . `docker kill -s HUP tileserver-gl` did work for me once I gave my run a name.

Though reading it I feel like this was originally supposed to be more of a refresh, but it seems like a kill... (maybe so docker will automatically refresh?)

zstadler:
@trondat ,

Please reload the configuration using the method described in [the documentation](https://tileserver.readthedocs.io/en/latest/usage.html#reloading-the-configuration):

> Reloading the configuration
> ======
> 
> It is possible to reload the configuration file without restarting the whole process by sending a SIGHUP signal to the node process.
> 
> - The `docker kill -s HUP tileserver-gl` command can be used when running the tileserver-gl docker container.
> - The `docker-compose kill -s HUP tileserver-gl-service-name` can be used when tileserver-gl is run as a docker-compose service.

The `docker-entrypoint.sh` shell script and [the documentation](https://github.com/maptiler/tileserver-gl/pull/560/files#diff-c4b284a0e8c330264ebefb4d3d14dd5bb68fe3b72c2ae656f49132e2c134b84f) were updated in #560 and #575 to avoid the need for the complex `docker exec` command required earlier:

```
- When running the tileserver-gl docker container, the signal must be sent from within the container:
- ::

[...]

PR #630, opened 2022-10-27
Fix HUP kill not reloading config

acalcutt:
This is a fix for https://github.com/maptiler/tileserver-gl/issues/629, which was broken in the syntax change of https://github.com/maptiler/tileserver-gl/pull/606

After looking at the code changes, it seems like the sections of code that handles SIGINT/SIGHUP and restarting was not being used anymore. This change renames the functions a bit so the SIGINT/SIGHUP code is actually used.


zstadler:
I have tested this PR under Docker. Reloading of an updated config file works in both ways:
- The documented preferred way: `docker kill -s HUP tileserver-gl`
- The older way: `docker exec tileserver-gl bash -c 'kill -HUP $(ls -l /proc/*/exe | sed -n "/\/node$/s/.*proc\/\([0-9]\+\)\/exe .*/\1/p")`

mnutt (approved):
I suspect that clearing the require cache might have been to give the ability to reload `options.dataDecorator` which is loaded via `require()`. IMO that seems like it's approaching ad-hoc hot-reloading territory and I think it's fine to not support.

zstadler:
Please consider resolving #629

acalcutt:
I'm planning to resolve it after the docker is updated

PR #632, opened 2022-10-28
Fix 'use omit=dev' warning. copy package-lock.json if it exists

acalcutt:
Changes "--production" to "--omit=dev" to avoid docker build warning.

Adds a package-lock.json to the project.. Copy package-lock.json to docker it exists.


mnutt (approved):
This looks good, I think instead of `npm install` we could also try using `npm ci` which ensures that it uses only the exact version specified in the package-lock.json?

PR #634, opened 2022-11-02
chown /data volume in Dockerfiles

nbr23:
Currently, if we run the container without mounting a volume, it fails with the following as /data is owned by root:

```
docker run --rm --pull always maptiler/tileserver-gl
latest: Pulling from maptiler/tileserver-gl
Digest: sha256:6bc09421fca44ef603ba174eb7b3e988464a73423ec56126fd3e771d4db370a8
Status: Image is up to date for maptiler/tileserver-gl:latest
Starting tileserver-gl v4.1.2
No MBTiles found
[DEMO] Downloading sample data (zurich_switzerland.mbtiles) from https://github.com/maptiler/tileserver-gl/releases/download/v1.3.0/zurich_switzerland.mbtiles
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: EACCES: permission denied, open 'zurich_switzerland.mbtiles'
Emitted 'error' event on WriteStream instance at:
    at WriteStream.onerror (node:internal/streams/legacy:62:12)
    at WriteStream.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: 'zurich_switzerland.mbtiles'
}
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,7 @@ COPY --from=builder /usr/src/app /usr/src/app
 
 COPY . /usr/src/app
 
+RUN mkdir -p /data && chown node:node /data
 VOLUME /data
 WORKDIR /data
 
--- a/Dockerfile_light
+++ b/Dockerfile_light
@@ -21,6 +21,8 @@ RUN set -ex; \
     rm -rf /var/lib/apt/lists/*;
 
 EXPOSE 80
+
+RUN mkdir -p /data && chown node:node /data
 VOLUME /data
 WORKDIR /data
 ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]


acalcutt:
I'm not sure about this one, I always considered it a bit of a security feature of running in the docker that it could not write back unless it was running docker as the same 'node' user. The documentation has recommended downloading the mbtiles file first to avoid this issue when running docker as a user that is not 'node'

While this would make it work without an error, I'm not sure node should always be able to write back to your data directory.

Anyone else have any opinions on this? am I thinking about it wrong?

scara:
Hi @acalcutt,
to me it looks like a clean-up useful for testing purposes: in real life when you'll mount an external volume in K8s it won't hurt nor it will solve permissions issues. Same applies w/ Docker Engine using an external folder.

HTH,
Matteo

zstadler:
Adding the following line to `Dockerfile` resolves the issue:
```
RUN mkdir -p /data; chmod a+w /data
```
This ensures that `/data` exists and is writable even when the image is run without mounting an external volume on it.

zstadler:
I think this PR is safe. The created `/data` and its change in ownership is limited in 2 ways:
1. It refers to the `/data` mount point in the image
2. It is applicable when nothing was mounted there

For example, if an empty read-only directory is mounted at `/data` the image created by this PR will not be able to write the required `zurich_switzerland.mbtiles` file in it, and the execution fails. Needless to say that nothing is written to the host's directory.

I believe this PR should be merged.

acalcutt (approved):

PR #637, opened 2022-11-06
Add .gitattribute

zstadler:
Lines in `.sh` scripts must always end with `lf`, even in Windows' git shell.

This PR follows @acalcutt's [request](https://github.com/maptiler/tileserver-gl/pull/636#issuecomment-1304603056) at the #636 discussion.


acalcutt (approved):
seems like a good idea to me, i've run into issues in other applications where commiting in windows messed up line endings.

Issue #323, opened 2018-10-18
serve multiple mbtiles with docker


Arturiko:
Hi,
So I'm using docker-compose for my maps server,
I downloaded two file of mbtiles and

**My docker-compose file is :**
version: '3.4'
services:
  openmaptiles:
    image: klokantech/tileserver-gl
    ports:
      - 8080:80
    volumes:
      - "./data:/data"
    command: "--verbose -c config.json" 

**My config.json file is :**

{
	"options": {
		"paths": {
			"root": "/usr/src/app/node_modules/tileserver-gl-styles",
			"fonts": "fonts",
			"styles": "styles",
			"mbtiles": "/data"
		}
	},
	"styles": {
		"klokantech-basic": {
			"style": "klokantech-basic/style.json",
			"tilejson": {
				"bounds": [ 34.07929, 29.37711, 35.91531, 33.35091 ]
			}
		},
		"osm-bright": {
			"style": "osm-bright/style.json",
			"tilejson": {
				"bounds": [ 34.07929, 29.37711, 35.91531, 33.35091 ]
			}
		},
		"klokantech-basic-gb": {
			"style": "klokantech-basic/style.json",
			"tilejson": {
				"bounds": [ -9.408655, 49.00443, 2.25, 61.13564 ]
			}
		},
		"osm-bright-gb": {
			"style": "osm-bright/style.json",
			"tilejson": {
				"bounds": [ -9.408655, 49.00443, 2.25, 61.13564 ]
			}
		}
	},
	"sources": {
		"israel": {
			"url": "mbtiles://{v2}",
			"type": "vector"
		},
		"greatBritain": {
			"url": "mbtiles://{v3}",
			"type": "vector"
		}
	},
	"data": {
		"v2": {
			"mbtiles": "2017-07-03_asia_israel-and-palestine.mbtiles"
		},
		"v3": {
			"mbtiles": "2017-07-03_europe_great-britain.mbtiles"
		}
	}
}

But on server I can see only one of them, like, randomly one of them :
![server](https://image.ibb.co/iKRfi0/tile-server.png)

As you can see, now it serves only GB map and other map does not exists.

what do i do wrong ?

**Thank you !**

spatialillusions:
A bit unclear what you expect here. The four styles you have in your config are displayed as styles, and the two datasets are displayed under data. If you want to have separate maps over Israel, you will have to copy the styles and change them to use the {v2} source.

Arturiko:
[...]

PR #645, opened 2022-11-23
feat: add `--watch` to reload on MBTiles change

erik:
This PR adds a new command line option, `-w, --watch`, which adds a file watcher to the MBTiles input (similar to the existing watcher on styles). It's a rough implementation honestly (requires restarting the express server), but seems to work.

The PR looks larger than it actually is - existing code was split out into helper functions, but the only real change is the additional `chokidar` object around line 190.


acalcutt:
Just so you know this is still something I plan to review, I just haven;t had the chance yet. 

anyone have any opinions on this? I think it sounds like a good feature, I just need to make some time to look at it better.

erik:
No problem! I've been using this daily for the past couple weeks as I've been working with Planetiler. So far so good. Main annoyance I'd call out is that you'll see (harmless) error messages printed to the console if the restart fails while the MBtiles file is actively being written to. The old version of the file is still served, so it's fine, just a bit noisy

src/main.js
```diff
@@ -83,90 +176,38 @@ const startWithMBTiles = (mbtilesFile) => {
...
+    })
+    .then((config) => startServer(null, config));
+
+  if (opts.watch) {
+    console.log(`[INFO] watching ${mbtilesFile} for changes...`);
+
+    chokidar
+      .watch(mbtilesFile, {
+        ignoreInitial: true,
+        // The file can be written to many times while tile generation is
```
scara, line 192: Typo here?
`to` insteaf of `too`.

HTH,
Matteo

scara:
Hi @acalcutt,
> I think it sounds like a good feature

Me too: we should focus on the file watching library to understand how it deals with external volumes, thinking at those running `tileserver-gl` as a container, especially those ones running it on K8s w/ replicas `>=2` and using `RWX` PVCs - the image should not store the data too, to avoid issues in distributing that big image in different nodes.

I do not have any experience with [`chokidar`](https://github.com/paulmillr/chokidar#how):
> Be advised that chokidar will initiate watchers recursively for everything within scope of the paths that have been specified, so be judicious about not wasting system resources by watching much more than needed.

I'd investigate in the initial use case too since I guess that the _MBTiles_ file should be presented to the server only when it is ready, through a copy, and not using that file as the working target file of the process charged to make changes on it or, if that is the case, to better understand what the changing process actually does when live-updating the tiles within the same file used by the server.
[`usePolling: true`](https://github.com/paulmillr/chokidar#performance) seems sensitive here when the share is network based like an `RWX` is.

I'd also work on [the events](https://github.com/paulmillr/chokidar#api) we want to listen to: we could probably avoid [`all`](https://github.com/maptiler/tileserver-gl/pull/645/files#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79R199) but I'm not an expert here; my concern is to avoid server reloading when not required.

Why not even add a filter on the extension e.g. `*.mbtiles`?

Guessing that on stopping the server we should also stop watching the target files.

Finally, how could this cope with multiple MBTiles files e.g. https://github.com/maptiler/tileserver-gl/issues/323#issue-371687861?

HTH,
Matteo

src/main.js
erik: No, this is meant as "to", as in "writing to a file"

erik:
Adding some additional context to help with review.

> I do not have any experience with [`chokidar`](https://github.com/paulmillr/chokidar#how):
> 
> > Be advised that chokidar will initiate watchers recursively for everything within scope of the paths that have been specified, so be judicious about not wasting system resources by watching much more than needed.

Only a path to a single provided is given to `chokidar` here.

> I'd investigate in the initial use case too since I guess that the _MBTiles_ file should be presented to the server only when it is ready, through a copy, and not using that file as the working target file of the process charged to make changes on it or...

At least in my case, this is for development purposes (iterating on the tileset locally), rather than a production server. I suppose it I could imagine it in a production environment as well, but in that case restarting the server manually as part of a deployment doesn't seem unreasonable.

> I'd also work on [the events](https://github.com/paulmillr/chokidar#api) we want to listen to: we could probably avoid [`all`](https://github.com/maptiler/tileserver-gl/pull/645/files#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79R199) but I'm not an expert here; my concern is to avoid server reloading when not required.

Could be scoped down to `add` and `change` event handlers, yeah. Server won't reload unless the MBTiles file is valid though.

> Why not even add a filter on the extension e.g. `*.mbtiles`?

Watcher is for a single file. Filter would be redundant.

> Guessing that on stopping the server we should also stop watching the target files.

`SIGINT` is already handled by running `process.exit(1)`, which will kill the watcher. 

> Finally, how could this cope with multiple MBTiles files e.g. [#323 (comment)](https://github.com/maptiler/tileserver-gl/issues/323#issue-371687861)?

This I don't know, would need some direction from someone closer to the project.

scara:
TNX @erik for the detailed reply!
Hoping that will help @acalcutt and other key people to understand the value and the scope of this PR.

Matteo

acalcutt:
Is this only useable when you are loading a single mbtile file? if you use a config file does watch do anything?

erik:
Honestly I'm not very familiar with all the functionality here, so I wasn't aware there was another way to use it than a single file passed on the command line. So probably won't work if that's a separate codepath

acalcutt:
When it starts with config file it looks like it just passes the config to startServer  ( https://github.com/maptiler/tileserver-gl/blob/master/src/main.js#L206 )

In the config there could be multipe mbtiles set like https://github.com/acalcutt/wifidb-tileserver-gl/blob/master/tileserver-gl/configs/config-3.1.1.json

I would love to see this work with multiple mbtiles in the config. In my project I update several files a night and have to restart tileserver to make them update. I would love just to set watch and be able to monitor all these mbtiles files.

erik:
Yeah, definitely won't support config files as-is then. I can take a look at adding support for that in this PR when I get a chance. Feel free to take over the PR if you'd like, not sure how soon I'll be able to work on it

erik:
Hey @acalcutt - gave this PR another look and don't think I'm the right person to implement it unfortunately. The different layers of the code involved here are a little too tightly coupled for me given my unfamiliarity with the codebase. I'll close this PR since the code we have here would need to be almost fully redone to support config files anyway. Maybe a new issue should be opened instead to keep track of the feature idea. Thanks for reviewing!

acalcutt:
If you do ever get the urge to work on this again, I was wondering if the watching could be moved into server.js. or if that would not work because it would have to restart itself?

Maybe in server.js we could go through each of the files in [config.data](https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L160) and add a watcher for each mbtiles file in that array? Right now if you use a single mbtiles file, main.js basically writes it's own config file and passes it along to server.js. if a config already exists main.js just passes it directly to server.js. I would think that means either start method should have a config file with an array of data mbtiles files that could just be gone through.

If it couldn't be done in server.js, maybe in main you could go through the data files in the config used in these two spots
https://github.com/maptiler/tileserver-gl/blob/b9df1d29d16e5c6b6f6cd4c9b8f967037c7d2ff1/src/main.js#L206
https://github.com/maptiler/tileserver-gl/blob/b9df1d29d16e5c6b6f6cd4c9b8f967037c7d2ff1/src/main.js#L167
I would think both those config files should should both have a config['data'] array to pull the mbtiles out of and set you watchers on.

Issue #642, opened 2022-11-15
Reflected XSS in /fonts/


procogent:
There is a reflected cross-site scripting vulnerability for requests under /fonts/.  This has been verified against tileserver-gl v4.1.2.

To recreate, access a tileserver-gl instance using the following URL (replace "host" with your tileserver-gl server):
http://host:8080/fonts/%3Cimg%20src%3da%20onerror%3dalert(1)%3E/0-255.pbf

You should see a popup demonstrating the XSS.

acalcutt:
How would this be used in real life? I feel like this would have to be included in a style to be loaded right? So problem when using ServeAllStyles ?

procogent:
@acalcutt A perpetrator could use this to create a malicious link, which they would embed into an email or third party website.  The link could be embedded inside an anchor text that provokes the user to click on it, which initiates the XSS request to the exploited website, reflecting the attack back to the user.  If tileserver-gl is being hosted on a domain with a login, for example, it could be used to steal the visitor's session cookies.

PR #647, opened 2022-11-24
fix: Ensure font serving does not reflect userdata-derived errors as HTML

mnutt:
Fixes https://github.com/maptiler/tileserver-gl/issues/642.

The issue is that if we could not find a font stack, we were rejecting with an error that included the passed fontstack name. We then sent that error back to the user as plaintext, but in not setting a `content-type` header, browsers guess the response is html and would try to render the fontstack name as html.

I audited the rest of the codebase to find other places where we might potentially reflect back user-supplied data and changed those to always set `content-type: text/plain`. I'd also be amenable to _always_ setting `content-type: text/plain` for any response that is not json and is not explicitly trying to render some other format.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tileserver-gl",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "tileserver-gl",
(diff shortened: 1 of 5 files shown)


acalcutt (approved):
Tested this and it seems to work. Before the change I get the popup mentioned
image

After the change, the popup no longer happens and the text is just displayed.
image

Looks good to me.

Issue #646, opened 2022-11-23
TileServer GL : issue with port 8080


raphael10-collab:
This is the current situation : 

O.S.: Ubuntu 22.04
Server : VPS Contabo Server

    root@vmi1046457:~# sudo docker run --restart=always -it -d -v /root/openmaptiles/data:/data -p 8080:80 maptiler/tileserver-gl
    9cfafc58933b3e8080c27b92463d276d1c46b718d28a59665a8cb5a945ebb905

    root@vmi1046457:~# docker ps
    CONTAINER ID   IMAGE                    COMMAND                  CREATED       STATUS                         PORTS                            NAMES
    9cfafc58933b   maptiler/tileserver-gl   "/usr/src/app/docker…"   2 hours ago   Up 2    
    hours (healthy)   8080/tcp, 0.0.0.0:8080->80/tcp   dazzling_colden

    root@vmi1046457:~# sudo ufw allow 8080/tcp
    Skipping adding existing rule
    Skipping adding existing rule (v6)

    root@vmi1046457:~# nc -l -p 8080 
    nc: Address already in use
    root@vmi1046457:~# 

     root@vmi1046457:~# netstat -an|grep LIST 
    tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN     
    tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN     
    tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN     
    tcp        0      0 127.0.0.1:5432          0.0.0.0:*               LISTEN     
    tcp        0      0 127.0.0.1:42069         0.0.0.0:*               LISTEN     
    tcp6       0      0 :::22                   :::*                    LISTEN     
    tcp6       0      0 :::80                   :::*                    LISTEN     
    unix  2      [ ACC ]     STREAM     LISTENING     722894   /run/user/0/systemd/private
    unix  2      [ ACC ]     STREAM     LISTENING     722900   /run/user/0/bus
    unix  2      [ ACC ]     STREAM     LISTENING     722902   /run/user/0/gnupg/S.dirmngr
    unix  2      [ ACC ]     STREAM     LISTENING     722904   /run/user/0/gnupg/S.gpg-agent.browser
    unix  2      [ ACC ]     STREAM     LISTENING     722906   /run/user/0/gnupg/S.gpg-agent.extra
[...]

PR #651, opened 2022-12-04
usage.rst: Add "Docker and `--port`" section

zstadler:


acalcutt:
Changing the port in the docker was mentioned in https://github.com/maptiler/tileserver-gl/issues/646#issuecomment-1326783419 , and at the time I was thinking that would break the help check. This is probably a good idea.

acalcutt (approved):
I'm going to approve this. one thing to note is the doc website doesn't seem to be updating and I have no way to push changes to it... so that is unfortunate.

zstadler:
> I'm going to approve this. one thing to note is the doc website doesn't seem to be updating and I have no way to push changes to it... so that is unfortunate.

The previous change to usage.rst is shown on the docs site. According to their docs, it is supposed to be done automatically. Let's wait a few days and see.

PR #719, opened 2023-01-18
chore: remove unused `run.sh` 🧽

vinayakkulkarni:
- [x] Removes the `run.sh` which was previously used in Dockerfile(s) but now the `docker-entrypoint.sh` is being used and this file was missed while refactoring of Dockerfile(s)


acalcutt:
I'm not aware of anywhere run.sh is used anymore. I think it's a good idea to remove it.

Is anyone else aware of where this may still be used?

acalcutt (approved):

PR #721, opened 2023-01-19
chore: remove {{public_url}} from static assets in favor of relative paths

mnutt:
Note: I'm not expecting this to be immediately merged, just starting a discussion around if this is something we want.

Continuing in my quest to reduce our chances of XSS by removing interpolation from our html templates, this PR removes the need for `{{public_url}}` to refer to static assets. It also places static assets under `assets/` to prevent future namespace collisions between static assets and routes.

One caveat: if you mount your app at `/foobar` and load `http://localhost:8080/foobar`, the relative URLs _won't_ work. The index page works around this by auto-redirecting to `http://localhost:8080/foobar/` if you load `http://localhost:8080/foobar`. This _seems_ acceptable to me as a common thing submounted pages might do but I definitely wanted to get feedback on it.

### DIFF
--- a/public/resources/index.css
+++ b/public/resources/index.css
@@ -1,18 +1,18 @@
 @font-face {
   font-family: 'OpenSans';
-  src: url('/fonts/OpenSans-Regular.ttf');
+  src: url('./fonts/OpenSans-Regular.ttf');
   font-weight: normal;
   font-style: normal;
 }
 @font-face {
   font-family: 'OpenSans';
-  src: url('/fonts/OpenSans-Italic.ttf');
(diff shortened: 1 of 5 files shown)


acalcutt:
I've seen a lot of PRs relating to public_url, so it leads me to believe it is used by people out there....not sure I'd want to get rid of it.

mnutt:
It's retained everywhere we absolutely need it, namely within JSON responses that require absolute URLs. My expectation is that everything in this PR should work for anyone using `public_url`.

My thought (and feel free to disagree) is I think we can/should eventually get to the entire frontend being static html.

PR #722, opened 2023-01-19
Remove IE support

acalcutt:
Remove Internet explorer support from view and data templates.


acalcutt:
Hmm, one strange thing I notice is the text looks off. leaflet and maplibre are 1 zoom level off, but even if you account for that the labels look way overzoomed in the mapblibre version of the raster. I find that strange since I though ill the rendering took place in maplibre-native...

Leaflet version (raster)
image

Maplibre version (raster)
image

acalcutt:
I figured it out based on https://github.com/maplibre/maplibre-gl-js/issues/1257

It was because I wasn't specifying a tileSize, so maplibre assumes a 512x512px tile. But tileserver is making a 256x256px tile. If you specify tileSize, maplibre downloads tiles at zoom level + 1 and the text looks better.

mnutt (approved):

PR #767, opened 2023-02-16
Simplify signal handling

zstadler:
Avoid implementing a complex signal relaying mechanism in `docker-entrypoint*.sh`.
Instead, the shell scripts perform `exec node ...` and the `node` process receives the signals directly from Docker

Related to #560, #575, which were my initial implementation.

PR #782, opened 2023-02-28
Handle SIGTERM

### LINKED ISSUE #781: Termination signal not being handled when deployed on Kubernetes since v4.4.5
*Problem*
I am deploying Tileserver GL Light on a Kubernetes cluster using the `maptiler/tileserver-gl-light` image. When I was using v4.4.4 release (and earlier), my pods responded appropriately to termination signals when Kubernetes was trying to delete them. I could see log lines like `Caught signal 15, stopping gracefully` and each pod would die quickly.

After upgrading to v4.4.5, the pods no longer respond to termination signals. Kubernetes waits the standard 30s grace period before force killing them. No log lines appear indicating that a signal has been caught and shutdown was graceful.

Is this related to https://github.com/maptiler/tileserver-gl/pull/767, which was released in 4.4.5?

zstadler:
Resolve #781

Testing:

```
$ docker run --detach -v $(pwd):/data -p 8080:80 --name tileserver-gl maptiler/tileserver-gl-l
ight:local --mbtiles zurich_switzerland.mbtiles
544505f228138f02fdb0a99b2c44500948af4fb061ea86e68e7ba6c98a3c189c

$ docker ps --filter "ancestor=maptiler/tileserver-gl-light:local"
CONTAINER ID   IMAGE                                COMMAND                  CREATED         STATUS                            PORTS                            NAMES
544505f22813   maptiler/tileserver-gl-light:local   "/usr/src/app/docker…"   7 seconds ago   Up 6 seconds (health: starting)   8080/tcp, 0.0.0.0:8080->80/tcp   tileserver-gl

$ docker kill --signal HUP tileserver-gl
tileserver-gl

$ docker kill --signal TERM tileserver-gl
tileserver-gl

$ docker ps --filter "ancestor=maptiler/tileserver-gl-light:local"
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES

$ docker logs tileserver-gl
Starting tileserver-gl v4.4.8
[INFO] Automatically creating config file for zurich_switzerland.mbtiles
[INFO] Only a basic preview style will be used.
[INFO] See documentation to learn how to create config.json file.
Run with --verbose to see the config file here.
Starting server
Listening at http://[::]:8080/
Startup complete
Stopping server and reloading config
Starting server
Listening at http://[::]:8080/
Startup complete
GET /health 200 2 - 3.961 ms
```

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -606,6 +606,10 @@ export function server(opts) {
     process.exit();
   });
 
+  process.on('SIGTERM', () => {
+    process.exit();
+  });
+
   process.on('SIGHUP', () => {
     console.log('Stopping server and reloading config');
 


acalcutt:
I assume the process ends after "GET /health 200 2 - 3.961 ms"?

Can you also add a console.log message, Could you add a console.log in there. Maybe just "Stopping server" to match the other message for SIGHUP.

I was also wondering about SIGINT/SIGTERM being the same. Should they have a common funtion like https://stackoverflow.com/a/66637892 ?

zstadler:
> I assume the process ends after "GET /health 200 2 - 3.961 ms"? Could you add a console.log in there. Maybe just "Stopping server" to match the other message for SIGHUP.

- Yes, the container was terminated after the heathcheck.
- I find it odd to add this message
  - The issue requested that SIGTERM have the same behavior as before.
  - The SIGHUP message is there in order to log the refresh of the server configuration.

However, if you feel strongly about it, please feel free to add it.

> I was also wondering about SIGINT/SIGTERM being the same. Should they have a common function like https://stackoverflow.com/a/66637892 ?

That would make sense if there was anything to do rather that `process.exit()`. Currently, `process.exit()` serves as the common function.

acalcutt:
`The issue requested that SIGTERM have the same behavior as before.`

When SIGTERM happened before, the old entrypoint file would display ["Caught signal 15, stopping gracefully"](https://github.com/maptiler/tileserver-gl/blob/c134795b8113b4bb676e0237ce4a681d6097f67a/docker-entrypoint.sh#L7) . I think it is good to show the process is ending gracefully (for example I couldn't tell from your log the SIGTERM was called. If i was reviwing that it would be good to know).

zstadler:
I now see that the removed shell script printed these messages, in this section of the code which is now removed:

https://github.com/maptiler/tileserver-gl/blob/c134795b8113b4bb676e0237ce4a681d6097f67a/docker-entrypoint.sh#L5-L9

This makes me wonder how #781 was introduced in the first place. As far as I can tell, when the shell script received a signal, it printed the message and sent the same signal to the node process. After #767, the signals are sent directly to the node process, but a received SIGTERM does not seem to have any impact.

Regardless of this mystery, I've added `Caught signal ...` messages to the code.

Note that the numeric signal values, as used in the shell version, were replaced by signal names, as used in nodeJS.

zstadler:
@acalcutt ,

Could you check why **The Pipeline / CI / ci (pull_request)** test fails with the following error, while it passed before:

> Error: This action does not have permission to create annotations on forks. You may want to run it only on `pull_request_target` events with checks permissions set to write. See https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions for details.

acalcutt:
It looks like there was a link issue. Locally I ran 'npm run lint:js', then 'npm run lint:js:fix'

It looks like it was mostly indentation (4 spaces vs 2, and and extra line)

acalcutt (approved):
I tested this with the commands above it seems to work as expected. Thanks for making those changes and fixing this.

```
# docker logs test10
Starting tileserver-gl v4.4.8
No MBTiles specified, using zurich_switzerland.mbtiles
[INFO] Automatically creating config file for zurich_switzerland.mbtiles
[INFO] Only a basic preview style will be used.
[INFO] See documentation to learn how to create config.json file.
Run with --verbose to see the config file here.
Starting server
Listening at http://[::]:8080/
Startup complete
Caught signal SIGHUP, refreshing
Stopping server and reloading config
Starting server
Listening at http://[::]:8080/
Startup complete
GET /health 200 2 - 15.981 ms
Caught signal SIGTERM, stopping gracefully
```

acalcutt:
> This makes me wonder how #781 was introduced in the first place. As far as I can tell, when the shell script received a signal, it printed the message and sent the same signal to the node process. After #767, the signals are sent directly to the node process, but a received SIGTERM does not seem to have any impact.

That's a good question. maybe the message gave the illusion of working before and the process was just killed anyway? 

Either way, it seems like it should handle it now. Thanks.

zstadler:
> That's a good question. maybe the message gave the illusion of working before and the process was just killed anyway?

 It was actually working, I tested it by sending SIGTERM manually only once.

> Either way, it seems like it should handle it now. 

Agreed. I was hoping maybe someone could solve this puzzle for me...

Issue #38, opened 2016-08-08
Support V2 spec tiles


klokan:
The OSM2VectorTiles project has upgraded recently to V2 specification of vector tile format - so all the vector tiles downloaded after OSM2VectorTiles milestone v2.2 are in this format.

We should upgrade the "MapBox GL Native" library too as there are reported rendering issues probably related to incompatibility of the old version of the rendering library with the new V2 spec of vector tiles. (to be reproduced first, before committing the upgraded version)

klokan:
Report from @daniyel:

Today I downloaded new switzerland.mbtiles, it looks like I am getting only 200 and 304 which is good I assume but I am still getting this funny rendering somehow:

![meilen_tile](https://cloud.githubusercontent.com/assets/356247/17476259/5c6c7b94-5d5f-11e6-99fd-256becc95f41.jpg)
![rapperswill_tile](https://cloud.githubusercontent.com/assets/356247/17476260/5c6d7bb6-5d5f-11e6-864f-7f3d429ab6d9.jpg)
![swiss_part](https://cloud.githubusercontent.com/assets/356247/17476261/5c6dbc0c-5d5f-11e6-8dba-5bc1804fd41e.jpg)

petrsloup:
I'm having hard time reproducing the issue -- I downloaded latest Switzerland and Zurich extracts, everything seems to be working ok (both raster and vector viewer). I even tried Paraguay (from https://osm2vectortiles-downloads.os.zhdk.cloud.switch.ch/v2.0/extracts/paraguay.mbtiles), which seems to be the newest file on the server (`2016-08-07T14:18:58.000Z`), but no visual errors.

Anyway, I updated the dependencies in a separate branch (`v2_tiles`).

@daniyel Where did you download the file? Are you having problems with the "Vector" or the "Raster" viewer? Can you try running `docker run -it -v $(pwd):/data -p 8080:80 klokantech/tileserver-gl:v2_tiles` to see if it helps?

klokan:
@daniyel we need your feedback.

daniyel:
Sorry for not answering earlier. I have downloaded the tiles from `https://osm2vectortiles-downloads.os.zhdk.cloud.switch.ch/v2.0/extracts/switzerland.mbtiles` and started the docker with `docker run -it -v $(pwd):/data -p 8080:80 klokantech/tileserver-gl:v2_tiles`. It seems the rendering is working now, atleast I could not see any distortions in the map.

[...]

PR #822, opened 2023-04-03
Windows support workaround.

acalcutt:
This is a workaround to allow tileserver-gl to run on windows.

Based on information from [tdcosta100](https://github.com/tdcosta100) about a sharp/canvas incompatibility on windows being caused by a MinGW64/UCRT64 build difference, I was able to work with the node-canvas people to get them to change their windows binary to a [UCRT64 build](https://github.com/Automattic/node-canvas/pull/2219). With this change, canvas 2.11.2 now can work with sharp 0.32.0

There is one other difficulty, which is the order sharp and canvas get imported matters, and unfortunately the order on windows is the opposite of what linux needs (what we have been using). we also can't just change the order of the imports with an if statement, because imports are required to be at the top level.

This PR works around the issue by separating out out the canvas/sharp imports into their own platform specific files. This allows the order to be changed per platform and allows tileserver-gl to run on windows (in addition to previously supported OSs)

With this change, tileserver-gl running on windows and linux!
image

image


scara:
HI @acalcutt,
great news!

What about adding an `INFO` log telling about the environment on which `tileserver-gl` will start?
I'm thinking about at least both the _OS_ and the _Architecture_.

HTH,
Matteo

acalcutt:
@scara I've updated the npm section of installation.rst to include windows.  I've also updated the layout a little and tried to make it a bit more accurate.

you can view the changes at
https://github.com/WifiDB/tileserver-gl/blob/windows_workaround/docs/installation.rst#native-dependencies

scara:
Hi @acalcutt,
> I've updated the npm section of installation.rst to include windows. 

Thanks!

I still wonder if we should add the "fingerprint" of the underline OS when starting, just for debug/support purposes.

HTH,
Matteo

acalcutt:
I think i misread what you were looking for the first time. I agree, I think os version and architecture are useful in the startup info. 

I added in some code to display os.versions and os.arch on the line with the tileserver version. I think that gives the best information (though it was hard to get exactly the version of the os like I wanted).

It looks like this now
image
image

scara:
Hi @acalcutt,
> I think i misread what you were looking for the first time

No problem.
Thanks for the new data in the log, even if `version()` looks "weird" when running on Linux: `platform()` could be better but we'll miss what is the distro behind that _Linux_ e.g. `#38~20.04.1-Ubuntu SMP Sun Mar 12 04:47:58 UTC 2023`.
On _Linux_ `os.version().split(' ').shift()` could help but it will break the _Windows_ prompt.

BTW guessing that even `arch` is a function i.e. it should be: `${os.arch()}`.

HTH,
Matteo

acalcutt:
This no longer seems to work with the latest version of sharp, so this is proably not the way to go.

acalcutt:
I am closing this since windows support is working in v4.7.0

PR #829, opened 2023-04-08
Better compatibility between canvas and sharp in Windows

tdcosta100:
Since canvas and sharp have known issues regarding compatibility in Windows, this PR addresses it installing the complete libraries in sharp directory (currently it uses a stripped down version of glib2 library, generating the infamous `The specified procedure could not be found.` error). This change does not affect Linux and Mac OS, only Windows.


acalcutt:
I tested this and it works. Seems like a good alternative to https://github.com/maptiler/tileserver-gl/pull/822 since we don't have to swap the order of imports to get it to work

Any opinions on fixing windows support this way?

acalcutt:
is there a reason you made that import script cjs. seems like it would easy to use esm format like the rest of the code (basically replacing the require with an import and changing the extension to js)

tdcosta100:
No special reason, it was the first way I tried. Just switched to .js, like the rest of the modules.

acalcutt (approved):

acalcutt:
@tdcosta100 could you run "npm run lint:js:fix" on this. i think that will vix the vague CI lint error.

tdcosta100:
@acalcutt, done!

PR #908, opened 2023-06-27
Resolve composite watermark issue

fvarga94:
Hi, I encountered a problem when using watermarks on static image API. When watermark is defined all other custom data on the static image is overridden.

I build and use the project using the available docker files, test data and config.

Only change is in the config.json where I added "watermark": "WATERMARK" under options.

Query:
/styles/test-style/static/auto/256x256@2x.png?path=8.519704,47.380481|8.519704,47.390481

Result:
image

Expected result:
image

I think this PR resolves the issue.


acalcutt (approved):
I tested this issue and fix and it seems to work as you described. 

Thanks for the fix.

PR #972, opened 2023-09-09
Static images: Allow styling each path differently

samuel-git:
Each path was rendered with each path style, drawing over the correctly styled path. Now each path gets only rendered once with the specific path style.

This is now possible:
750x250@2x

With this URL:
```
http://localhost:8080/styles/basic-preview/static/auto/750x250@2x.jpg?path=stroke:yellow|width:2|fill:green|5.9,45.8|5.9,47.8|10.5,47.8|10.5,45.8|5.9,45.8&path=stroke:blue|width:3|fill:red|4.9,44.8|4.9,46.8|9.5,46.8|9.5,44.8|4.9,44.8
```

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -442,9 +442,10 @@ const drawMarkers = async (ctx, markers, z) => {
  * @param {object} ctx Canvas context object.
  * @param {List[Number]} path List of coordinates.
  * @param {object} query Request query parameters.
+ * @param {string} pathQuery Path query parameter.
  * @param {number} z Map zoom level.
  */
-const drawPath = (ctx, path, query, z) => {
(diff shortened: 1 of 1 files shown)


src/serve_rendered.js
```diff
@@ -446,103 +446,100 @@ const drawMarkers = async (ctx, markers, z) => {
  * @param {number} z Map zoom level.
  */
 const drawPath = (ctx, path, query, pathQuery, z) => {
-  const renderPath = (splitPaths) => {
```
samuel-git, line 448: Inlined this function

acalcutt (approved):
I finally got a chance to test this and it seemed to work good. I tested with you example and examples from the previous PR and they all still seemed to work as expected.

 thanks for submitting this.

PR #993, opened 2023-09-26
fix: incorrect polyline decoding

### LINKED ISSUE #980: Static image : polyline being truncated when containing pipe character
When an encoded polyline contain a pipe character to generate a static image, it apparently truncate the given encoded polyline which result into an image not representing truthfully an encoded polyline.

Example, this polyline `{{biGwvyGoUi@s_A|{@` if used with following url : `http://localhost:8080/styles/osm-liberty/static/auto/1000x1000.jpg?stroke=red&width=3&path=enc:{{biGwvyGoUi@s_A|{@`
will create this image : 
image

but if you give the poyline to [GeoJSON ](https://geojson.io/#map=14.33/43.61005/1.43899) it will give this stroke : 

image

My guess is that tileserver truncate the polyline up to the pipe character and then dont take into account the second of the polyline which result in a way different stroke in generated image.

The fix would to not try splitting on pipe when an encoded polyline is given

boldtrn:
Fixes #980 

Don't split static image encoded path on pipe or comma. I t hink we can't support fill, stroke, and width in combination with encoded polyline, so we should remove it. The encoded polyline can contain all ASCI character, so it can contain `| ` or `,`, splitting on these characters destroys the polyline and will lead to incorrect render results.

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -35,13 +35,14 @@ Static images
 
 * All the static image endpoints additionally support following query parameters:
 
-  * ``path`` - comma-separated ``lng,lat``, pipe-separated pairs
-
-    * e.g. ``5.9,45.8|5.9,47.8|10.5,47.8|10.5,45.8|5.9,45.8``
-    * can be provided multiple times
-    * or pass the path as per `Maptiler Cloud API <https://docs.maptiler.com/cloud/api/static-maps/>`_
-    * Match pattern: ((fill|stroke|width)\:[^\|]+\|)*((enc:.+)|((-?\d+\.?\d*,-?\d+\.?\d*\|)+(-?\d+\.?\d*,-?\d+\.?\d*)))
-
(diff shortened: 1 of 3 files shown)


acalcutt (approved):

acalcutt:
Thanks for making the changes discussed in slack which still allow fill, stroke, and width in combination with encoded polyline.

Just a note to anyone who looks at this PR. we looked at what maptiler has done for this and it looks they have made it so when you specific 'enc:' then everything after is consider part of the polyline. 
`-e.g. _p~iF~ps|U_ulLnnqC_mqNvxq`@-If used, the rest of the path parameter is considered to be part of the encoded polyline string -- do not specify the coordinate pairs.`

With that in mind, this PR now changes the way the encoded part of the string is pulled from path so everything after 'enc:' is considered part of the polyline. this makes it so fill, stroke, and width can still exist before 'enc:', but not after.

PR #1005, opened 2023-10-05
Add support for stylized attribution text for static images

kochis:
Adds a new config option `staticAttributionText` that is used for generating the attribution in the static images API.

Open to any suggestions on the option naming, or styling choices. Tried to keep the style as close to default web attribution as possible.

Here's what the `staticAttributionText` looks like:

image

Here's what the default `watermark` looks like:

image

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -140,7 +140,13 @@ It is recommended to also use the ``serveAllFonts`` option when using this optio
 -----------
 
 Optional string to be rendered into the raster tiles (and static maps) as watermark (bottom-left corner).
-Can be used for hard-coding attributions etc. (can also be specified per-style).
+Not used by default.
+
+``staticAttributionText``
+-----------
+
+Optional string to be rendered in the static images endpoint. Text will be rendered in the bottom-right corner,
+and styled similar to attribution on web-based maps (text only, links not supported).
 Not used by default.
 
 ``allowRemoteMarkerIcons``
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -33,6 +33,7 @@ Example:
       "serveAllStyles": false,
       "serveStaticMaps": true,
       "allowRemoteMarkerIcons": true,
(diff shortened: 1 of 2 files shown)


acalcutt:
I like it. I do think attribution is important so having a way to add it is good. it would be nice if it came from the tilejson attribution, but i understand like you mentioned in slack that many attributions have html links that are hard to read as just text.

For name of the attribute i think i like `staticAttributionText` since this only applies to static images, unlike watermark which also applies to every tile.

The only other thing this needs is something in the docs, maybe we could put it around watermark https://github.com/maptiler/tileserver-gl/blob/master/docs/config.rst#watermark and possibly update the watermark text to remove the part about attribution.

kochis:
@acalcutt Took a stab at the suggestions and updated the docs.

acalcutt:
Thanks, it looks like this just needs lint now. I'm going to leave this up for a few days and see if anyone has any comments

acalcutt:
One thing to note, I know you changed this to only show in static mode so this would not longer show on all tiles in raster view. but I was thinking someone who is using the 'tilemargin' will likely see this on every tile still, because that option sets the map into static mode also.

kochis:
Ah, interesting about the `tileMargin` thing. I did set the `tileMargin` to a non-zero value to the tileMargin config (does 5 make sense as a value?) and didn't see the attribution show up on the raster maps. Unless this is expected to surface somewhere else.

kochis:
It looks like `opt_mode` is still tile when using `tileMargin`. It does use the static renderer, but doesn't change the `opt_mode`.

https://github.com/radarlabs/tileserver-gl/blob/296524a86420ab9d0773adb0b32927c8c90f0a2a/src/serve_rendered.js#L677-L681

acalcutt:
Nice, thanks for checking on tileMargin

acalcutt (approved):

PR #1023, opened 2023-10-18
Ensure ID is maintained over deconstructing object

### LINKED ISSUE #1022: style.json broken in 4.6.0
Hi, in https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L540

the code has changed from

```
style.id = id;
style.name = (serving.styles[id] || serving.rendered[id]).name;
style.serving_data = serving.styles[id];
style.serving_rendered = serving.rendered[id];
```

to

```
return {
      id,
      name: (serving.styles[id] || serving.rendered[id]).name,
      serving_data: serving.styles[id],
      serving_rendered: serving.rendered[id],
      ...style,
    };
```

which means the `id` value in `style` is overriding the parms set `id`. I'm not sure if this is intentional or not but has broken our setup as the `viewer` template is now returning the wrong `id` and returning 404 for the style.json

ryan-shaw:
Fixes #1022

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -546,11 +546,11 @@ function start(opts) {
     }
 
     return {
+      ...style,
       id,
       name: (serving.styles[id] || serving.rendered[id]).name,
       serving_data: serving.styles[id],
       serving_rendered: serving.rendered[id],
-      ...style,
     };
   });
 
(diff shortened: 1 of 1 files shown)


acalcutt (approved):
I tested this locally and it seems to work fine. Changing the order makes sense to me to better replicate the old functionality. Like you showed, id was getting overwritten before.

Issue #1025, opened 2023-10-20
4.5.2-4.6.1 breaks URL-based "glyphs" and "sprites" (concats URLs with root paths?)


alex-kowalczyk:
With the following config.json and mystyle.json (relevant fragments) 4.5.1 displays nice map,
while after updated to 4.6.0 it emits errors for glyphs and fonts and does not display vector map at all. Moreover, errors indicate that sprites URLs are concated with local "/data" root prefix. 
Revert to 4.5.1 fixed the issue.

Errors on startup:
```
mlgl: {
    class: 'Style',
    severity: 'ERROR',
    text: "Failed to load sprite: ENOENT: no such file or directory, open '/data/https://static.mydomain.eu/sprites/osm-liberty-topo@2x.json'"
}
```
Errors when serving:
```
ERROR: Trying to use Arial Unicode MS Regular as a fallback
ERROR: Font not found: Roboto Condensed Italic
ERROR: Trying to use Arial Unicode MS Regular as a fallback
ERROR: Font not found: Arial Unicode MS Regular
GET /fonts/Roboto%20Condensed%20Italic/0-255.pbf 400 41 - 4.819 ms
ERROR: Font not found: Arial Unicode MS Regular
```

config.json:
```json
{
  "options": {
    "paths": {
      "root": "/data",
      "styles": "styles",
      "fonts": "",
      "sprites": "",
      "mbtiles": ""
    },
    "domains": [
      "my.domain"
    ],
    "serveAllFonts": false,
    "serveStaticMaps": false
  },
  "styles": {
    "mystyle": {
      "style": "mystyle.json",
      "tilejson": {
        "type": "overlay",
        "bounds": [12.08477,48.54292,18.86321,51.06426]
      }
    }
  },
  "sources": {
     "europe": {
        "url": "mbtiles://{europe}",
        "type": "vector"
     }
  },
  "data": {
    "europe": {
      "mbtiles": "tiles.mbtiles"
    }
  }
}

```

mystyle.json:
```
{
  "version": 8,
  "name": "MyStyle",
  "center": [18.4803, 54.46639],
  "zoom": 13,
  "sources": {
    "mystyle": {
      "type": "vector",
      "url": "mbtiles://tiles.mbtiles"
    },
  },
  "id": "mystyle"
  "sprite": "https://static.mydomain.eu/sprites/osm-liberty-topo",
  "glyphs": "https://fonts.mydomain.eu/{fontstack}/{range}.pbf",
[...]

PR #1026, opened 2023-10-20
Fixed Issue #1025 - URL-based glyphs and fonts break displaying of ve…

alex-kowalczyk:
Fixed Issue #1025 - URL-based glyphs and fonts break displaying of vector tiles
    
**Root cause**: modified regex in #1002 caused URLs to be broken.
**Fixed**: by reverting previous regexes.
    
#### Tests:
 * Confirmed by building and running locally (via Dockerfile) with data folder which breaks on 4.6.0
 * Run tests via Dockerfile_test


acalcutt (approved):
I tested this locally and confirmed this issue. It looks like it started in v4.5.2 and is in the current v4.6.1 version.

I reverting these values and it seems to have fixed the issue described.

Issue #1006, opened 2023-10-06
Replace 'request'


acalcutt:
The request package used by tileserver-gl has been depreciated for quite a while and now has some vunerabilities. I think we should make some effort to replace it with something newer. I have looked a bit in the past and there are some options to replace it, but I have gotten stuck on understating the gzip features.

If anyone has any ideas to relace it I would love to see a PR. I think it only gets used in https://github.com/maptiler/tileserver-gl/blob/master/src/serve_rendered.js

mjj203:
I issued a pull request to your fork of tileserver-gl. Actually had to update main.js and serve_rendered.js with axios. https://github.com/acalcutt/tileserver-gl/pull/284

acalcutt:
Thanks, I looked at it and it looks good. Would you be able to make a PR on the maptiler/tileserver-gl master branch instead of my old fork? I have been putting in my effort here, so my fork is a bit out of date now.

PR #1043, opened 2023-11-02
Replace request with axios

acalcutt:
This is a cherry pick of the PR https://github.com/acalcutt/tileserver-gl/pull/284 submitted by @mjj203 in my old fork. I have updated it a little to fix some things changed in the current release.

This replaces request, which is unmaintained and has serveral vulnerabilies, with axios.

Fixes https://github.com/maptiler/tileserver-gl/issues/1006

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@maplibre/maplibre-gl-style-spec": "18.0.0",
         "@sindresorhus/fnv1a": "3.0.0",
         "advanced-pool": "0.3.3",
-        "axios": "1.5.1",
+        "axios": "^1.6.1",
         "canvas": "2.11.2",
         "chokidar": "3.5.3",
         "clone": "2.1.2",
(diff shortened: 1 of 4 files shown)


mnutt (approved):
This looks good to me. Great to see us replacing `request`.

`axios` is a good batteries-included http library. In my fork I used `undici`, which (as of node 16) will become the official node.js request library and uses web-standard `fetch`, but I think `axios` is also a good choice.

The code looks fine too. Down the road we might have an opportunity to refactor the maplibre data loading; maybe to something like:

```javascript
const renderer = new mbgl.Map({
  mode: mode,
  ratio: ratio,
  request: (req, callback) => {
    requestMap(req.url)
      .then((data) => callback(null, data))
      .catch((err) => callback(err, null));
});

function requestMap(url) {
  const protocol = url.split(":")[0];

  if (protocol === "sprites") {
    return requestSprites(url);
  } else if (protocol === "fonts") {
    return requestFonts(url);
  } else if (protocol === "mbtiles") {
    return requestMbtiles(url, dataDecoratorFunc);
  } else if (protocol === "http" || protocol === "https") {
    return requestUrl(url);
  } else {
    return Promise.reject(new Error(`Unkown protocol: ${protocol}`));
  }
}
```

This way we have the callback->promise transition fairly early on and can just deal in promises.

PR #1062, opened 2023-11-19
Extract nested functions and simplify a little

Caerbannog:
This is a first pass at simplifying the `server_rendered.js` file, which has grown quite big.

This PR will be easier to review commit by commit (there is a GitHub button for that).
If necessary I can split each commit in a dedicated PR.

There should have been no functional changes.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -1,0 +1,303 @@
+'use strict';
+
+import { createCanvas, Image } from 'canvas';
+
+import SphericalMercator from '@mapbox/sphericalmercator';
+
+const mercator = new SphericalMercator();
+
+/**
+ * Transforms coordinates to pixels.
+ * @param {List[Number]} ll Longitude/Latitude coordinate pair.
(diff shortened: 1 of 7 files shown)


src/render.js
```diff
@@ -0,0 +1,303 @@
...
+  ctx.strokeWidth = '1px';
+  ctx.strokeStyle = 'rgba(255,255,255,.4)';
+  ctx.strokeText(text, 5, height - 5);
+  ctx.fillStyle = 'rgba(0,0,0,.4)';
+  ctx.fillText(text, 5, height - 5);
+
+  return canvas;
+};
+
+export const rendeAttribution = (width, height, scale, text) => {
```
acalcutt, line 281: is this supposed to be "renderAttribution", look like the 'r' is missing

Caerbannog: Correct. Fixed.

acalcutt (approved):
I looked through these commits and tested them locally. I think these changes look good.

PR #1064, opened 2023-11-20
Add a gallery of tests to visually check for regressions

Caerbannog:
This is a new page that shows a collection of tests, mostly static maps, and the expected outputs.
For static maps it show the difference and computes a success/failure.

The files are served from the /public directory.
A special config is used to allow remote images and test copyrights and watermarks.
It expects the usual zurich mbtiles, but does not download it automatically.

## Instructions for use (also included)
Put zurich_switzerland.mbtiles in public/resources/gallery/ then run with:

docker build . && docker run --rm -it -p 8080:8080 $(docker build -q .) -V \
  --config /usr/src/app/public/resources/gallery/config.json

URL on tileserver: http://0.0.0.0:8080/gallery/gallery.html

## Screenshot of the page
Capture d’écran 2023-11-20 à 21 49 41


acalcutt:
I like the idea of this but I'm not about how it is being made its own page. 

This is similar to the maplibre-native render tests that get run as part of a github workflow. I would love to see something more like that where these image comparison tests could run in the workflow and use something like expect to compare the results, and then in the end have a report similar to the page you show that gets uploaded as part of the workflow.

PR #1075, opened 2023-11-23
Try to explain the zoom=0 special case

Caerbannog:
The code has a special case for zoom 0 that I did not understand at first.
Also, the comment that said `from 512 to 256` did not seem right since it assumes tiles are generated at `@1x` (when actually it can be more).

Please, dear reviewers, tell if it does not seem to you that these changes make it easier to understand.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -413,9 +413,8 @@ const respondImage = (
     pool = item.map.renderersStatic[scale];
   }
   pool.acquire((err, renderer) => {
-    const mlglZ = Math.max(0, z - 1);
     const params = {
-      zoom: mlglZ,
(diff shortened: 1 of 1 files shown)


acalcutt:
I'm not sure about this one. I think i liked it better before with a variable, though the comment could use some work.

I think the from 512 to 265 comment makes sense... The default image size generated by the maplibre-native rendering (with our settings) is 256px. At zoom level 0, maplibre-native doesn't support that level, so they basically make the image by requesting a 512px image from zoom level 1, then resizing it to the default 265px.

acalcutt:
This has been explained a bit in https://github.com/maptiler/tileserver-gl/pull/1152 which has been merged. I am closing this PR

PR #1076, opened 2023-11-24
Fix listFonts

Caerbannog:
Promise code never worked: listFonts did not wait for fs.stat() to resolve(). This was not evident because late results were still used.

A recent PR made it worse: late results are now ignored. This manifested for styles with missing fonts, no fallback could be used.


acalcutt (approved):
I tested this and it fixed the font issue I described in slack

PR #1077, opened 2023-11-24
Fix crash on ARM during rendering

Caerbannog:
The order of imports changed in PR #1062.

This causes a crash on ARM platforms during raster rendering.
To reproduce, load rendered image. The home page also triggers the crash through thumbnail rendering.

This PR fixes it, with a comment to minimize the chance of regression.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -18,7 +18,7 @@ import path from 'path';
 import url from 'url';
 import util from 'util';
 import zlib from 'zlib';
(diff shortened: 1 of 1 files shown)


acalcutt:
can you remove the comment on 
https://github.com/maptiler/tileserver-gl/blob/49445d89b7ee13613ba9daceea05d758158c3726/src/serve_rendered.js#L21
since that doesn't matter after this reordering

acalcutt (approved):

PR #1108, opened 2023-12-31
Hot-Reloading changed mbtiles

candux:
This will watch for changes in the mbtiles and hot-reload it.
To keep the implementation simple, the whole renderer is restarted.
This will create a small downtime.

It is very important to replace files atomically.

I tried to test all possible edge-cases, but we (geOps) don't use it in production yet. More testing would be much appreciated.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -32,6 +32,7 @@ Example:
       "pbfAlias": "pbf",
       "serveAllFonts": false,
       "serveAllStyles": false,
+      "watchMbtiles": false,
       "serveStaticMaps": true,
       "allowRemoteMarkerIcons": true,
       "allowInlineMarkerImages": true,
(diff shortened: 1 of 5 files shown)


acalcutt:
Nice, I think this will be a great feature. It looks like this is similar to https://github.com/maptiler/tileserver-gl/pull/455/files , so I am wondering how you feel this will be affected by the serveAllData, which was a concern in the other thread. I assume the files are just reloaded with the same same, so it wouldn't actually be much of an issue.

The other thing I notice is this focuses on mbtiles, but we now also support pmtiles as a Input File.  Can you update this to also support the pmtiles directory and input files that start with pmtiles:\\

candux:
> It looks like this is similar to https://github.com/maptiler/tileserver-gl/pull/455/files , so I am wondering how you feel this will be affected by the serveAllData, which was a concern in the other thread. I assume the files are just reloaded with the same same, so it wouldn't actually be much of an issue.

Yes, I don't think we have any of these problems in this MR. Now that I think of it, most of the concerns raised there should be already solved with this MR. When I started implementing this, I thought that a serveAllData option would be too difficult to get right.  But maybe it's not difficult anymore. I'll take another shot at implementing it. Otherwise, I'll update this PR with support for mbtiles

Elliott-Rose-BSC:
Any news on this feature?

PR #1136, opened 2024-01-19
Add support for 512 sized raster tiles  - v2

acalcutt:
This is my attempt to add 512px raster tile support to TileServer-GL. It is a continuation of https://github.com/maptiler/tileserver-gl/pull/495 , updated for the current branch

`Known issues`

1.) Zoom 0 in 512px raster mode does not work right. I think this does not work because in maplibre tile mode, the renderer is not able to render beyond a single 512x512px tile, but zoom  0 does this thing were it doubles the width height of the previous zoom level tile. In our default tiles this works fine, since in maplibre native the default tile size is 512, but we use smaller 256 size tiles, so doubling doesn't hit this restriction.

I was able to somewhat get around this by forcing zoom 0 to use the static render instead, but the scaling still didn't look quite right

2.) In data view I had no way to know if the source tiles are 256 or 512. I defaulted data view to 256px so it covers most tiles, but I think if you know your tiles are 512px you can set that tile size in your style sources and still get the benefit of the bigger tiles.

### DIFF
--- a/test/tiles_rendered.js
+++ b/test/tiles_rendered.js
@@ -1,5 +1,3 @@
-import { join } from 'path';
-
 var testTile = function (
   prefix,
   tileSize = 256,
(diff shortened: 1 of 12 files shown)


test/tiles_rendered.js
```diff
@@ -1,8 +1,30 @@
...
+  x,
+  y,
+  format,
+  status,
+  scale,
+  type,
+) {
   if (scale) y += '@' + scale + 'x';
-  const path = '/styles/' + prefix + '/' + z + '/' + x + '/' + y + '.' + format;
+  var path =
```
mnutt, line 13: this is fine as-is, if you wanted to you could use `require('path')` to join it all together:
```js
path.join('/styles', prefix, tileSize, z, x, `${y}.${format}`);
```

mnutt (approved):

PR #1149, opened 2024-01-26
Fix WMTS endpoint issues after 512px tile PR

acalcutt:
I found I had missed an issue in wmts in https://github.com/maptiler/tileserver-gl/pull/1136

The two issues were the following
1.) Both layers had the same identifier. This seemed to cause the wrong one to be loaded when I tested with arcgis online

2.) 512px tiles were misaligned, and getting 404 'invalid center' errors. I think this was because it was requesting the wrong zoom level tiles. I fixed this by raising the identifier by 1 for all 512pm TileMatixes

Before the wmts enpoint looked like this
image

But after these changes it looks correct, and I am able to search locations and get the correct place (where before i couldn't)
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tileserver-gl",
-  "version": "4.9.0",
+  "version": "4.9.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "tileserver-gl",
-      "version": "4.9.0",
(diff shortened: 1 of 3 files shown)


acalcutt:
One question I did have is what do the 'WGS84_512' and WGS84_256 TileMatrixSets actually do? They don't seem like they are used by the layers

mnutt (approved):

Issue #1187, opened 2024-02-26
Unhandled exception when pmtiles is used and GeoJSON is requested


Jollyfant:
In `serve_data.js` an exception is triggered because `isGzipped` is undefined: https://github.com/maptiler/tileserver-gl/blob/master/src/serve_data.js#L73 when `pmtiles` is used as data source and GeoJSON is requested.

Solution is to either add a check for Gzip magic bytes as per https://github.com/maptiler/tileserver-gl/blob/master/src/serve_data.js#L120 or remove the check entirely.

acalcutt:
Just wondering, do you have an example pmtiles geojson file I can test with? The PR looks good, I just realized geojson pmtiles isn't something I have tested before.

acalcutt:
Nevermind, I forgot this isn't a geojson formatted pmtiles, but a pbf pmtiles in geojson format. I can test with what I have.

acalcutt:
Thank for the PR, it has been merged. I will try to make a new release when I get home later tonight.

acalcutt:
Fixed by https://github.com/maptiler/tileserver-gl/pull/1188

PR #1188, opened 2024-02-26
Fix exception in pmtiles format when GeoJSON is requested

Jollyfant:
Fixes exception when format `pmtiles` is used and `GeoJSON` is requested. Here the variable `isGzipped` is always undefined and the check is redundant because the data is already unzipped. See #1187.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -69,12 +69,6 @@ export const serve_data = {
               headers['Content-Type'] = 'application/x-protobuf';
             } else if (format === 'geojson') {
               headers['Content-Type'] = 'application/json';
-
-              if (isGzipped) {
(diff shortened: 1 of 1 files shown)


acalcutt (approved):
Tested the changes and they fixed the error. 
image

Without this change geojson gave the error

```
file:///C:/Users/acalcutt/Documents/GitHub/tileserver-gl-maptiler/src/serve_data.js:73
              if (isGzipped) {
              ^

ReferenceError: isGzipped is not defined
    at file:///C:/Users/acalcutt/Documents/GitHub/tileserver-gl-maptiler/src/serve_data.js:73:15
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Issue #1268, opened 2024-06-04
Map Fonts: 400 Bad Request


vzkharov:
Locally all works great, even `Roboto` fonts are got.

Local Machine:
- System: Apple Silicon M2

But on Hetzner Server when fonts are requested — `400 Bad Request`

Production Setup:
- System: Ubuntu 20.4 LTS
- Logs:

```sh
GET /data/planet/3/4/3.pbf 200 439858 - 19.311 ms
GET /data/planet/4/8/4.pbf 200 175999 - 11.599 ms
GET /data/planet/4/8/5.pbf 200 375107 - 17.883 ms
GET /data/planet/4/9/4.pbf 200 193329 - 25.560 ms
GET /data/planet/4/9/5.pbf 200 314155 - 41.204 ms
GET /data/planet/4/9/6.pbf 200 345074 - 41.865 ms
GET /data/planet/4/8/6.pbf 200 303923 - 6.839 ms
GET /fonts/Roboto-Bold/7168-7423.pbf 400 2 - 1.494 ms
GET /fonts/Roboto-Bold/1024-1279.pbf 400 2 - 3.183 ms
GET /fonts/Roboto-Bold/256-511.pbf 400 2 - 1.753 ms
GET /fonts/Roboto-Regular/8192-8447.pbf 400 2 - 6.302 ms
GET /fonts/Roboto-Regular/7168-7423.pbf 400 2 - 6.767 ms
GET /fonts/Roboto-Regular/512-767.pbf 400 2 - 7.134 ms
GET /fonts/Roboto-Regular/1024-1279.pbf 400 2 - 10.770 ms
GET /fonts/Roboto-Regular/256-511.pbf 400 2 - 6.759 ms
GET /fonts/Roboto-Bold/65024-65279.pbf 400 2 - 5.670 ms
GET /fonts/Roboto-Regular/0-255.pbf 400 2 - 5.613 ms
GET /fonts/Roboto-Bold/4864-5119.pbf 400 2 - 2.697 ms
GET /fonts/Roboto-Bold/64256-64511.pbf 400 2 - 2.657 ms
GET /fonts/Roboto-Bold/4608-4863.pbf 400 2 - 4.108 ms
GET /fonts/Roboto-Bold/768-1023.pbf 400 2 - 2.11
```

(setup)
```sh 
Starting tileserver-gl-light v4.10.3
Using specified config file from config.json
Starting server
Listening at http://[::]:3001/
Style "basic" changed, updating...
[...]

PR #1294, opened 2024-06-28
feat: use `@jsse/pbfont` instead of `@mapbox/glyph-pbf-composite`

prashis:
- Use `@jsse/pbfont` instead of `@mapbox/glyph-pbf-composite`
- Fixes https://github.com/maptiler/tileserver-gl/issues/1268

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,7 +4,7 @@ import path from 'path';
 import fsPromises from 'fs/promises';
 import fs, { existsSync } from 'node:fs';
 import clone from 'clone';
-import * as glyphCompose from '@jsse/pbfont';
+import { combine } from '@jsse/pbfont';
 
 /**
  * Restrict user input to an allowed set of options.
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -4,7 +4,7 @@ import path from 'path';
 import fsPromises from 'fs/promises';
 import fs, { existsSync } from 'node:fs';
 import clone from 'clone';
-import glyphCompose from '@mapbox/glyph-pbf-composite';
+import * as glyphCompose from '@jsse/pbfont';
```
acalcutt, line 7: Can we just import the named combine function 'combine'

prashis: Yes, Have made the changes accordingly. :)

jessekrubin:
How performant is this tile server? just curious.

acalcutt:
I'd say performance it a tough thing to judge, and tileserver-gl has a lot of different configuration options that change performance. you can see an example of my server at https://tiles.wifidb.net , which is just npm running on a Ubuntu 22.04  virtual server with HA proxy in between, and a fallback server on a rasperry pie. the later pi is a bit slow rendering, but good for a backup/

Just a note on this PR. I am having a bit of issues with my permission on the repo, so it might take some time for me to get it fixed with maptiler and get this approved.

acalcutt (approved):

PR #1307, opened 2024-07-09
feat: added-cache-id

am2222:
This PR adds a `cache-id` query parameter to the tiles url. This helps to refresh client side cache when we update mbtiles on the server as its file hash changes and browser invalidates older caches

- if file is mbtile it uses fileHash
- if file is pmtiles it uses mtime (last modified time)

- I need some more information about the tests. On my machine I was not able to run tests as it was missing test data folder. is there any way I can access them?
Thanks

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "color": "4.2.3",
         "commander": "12.1.0",
         "cors": "2.8.5",
+        "crypto": "^1.0.1",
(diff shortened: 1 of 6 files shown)


acalcutt:
For the test data, there is a link to it here https://github.com/maptiler/tileserver-gl?tab=readme-ov-file#getting-started-with-linux-cli under "Test from command line"

PR #1320, opened 2024-07-18
fix: avoid synchronous calls where possible

akx:
These were a bunch of `*Sync` calls that were rather easy to convert to promisified ones, either via `util.promisify` or just `node:fs/promises`.

There's at least one more place I didn't touch, solely because it would have required refactoring all of `server.js::start` to be async.

There are other places left (out of scope of this PR) where e.g. callback async could be converted to promises, etc.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -3,6 +3,7 @@
 'use strict';
 
 import fs from 'node:fs';
+import fsp from 'node:fs/promises';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import axios from 'axios';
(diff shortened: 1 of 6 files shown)


acalcutt:
I would love another review of this, if anyone has the time. In general I think this is a good direction, but I am not really familiar. I assume this makes things run a bit more in parallel so it would increase response times?

akx:
> I assume this makes things run a bit more in parallel so it would increase response times?

It should decrease response times (that is, make things faster) where there are multiple concurrent requests to code paths which previously used `*Sync` calls (since a synchronous call will tie up the entire interpreter).

acalcutt (approved):
I tested this and it seems to work fine. seems like a good direction to get away from sycronous calls

PR #1321, opened 2024-07-18
refactor: add promisified wrapper for MBTiles

akx:
This adds a small wrapper around the `@mapbox/mbtiles` MBTiles class to promisify opening them (instead of callback pyramids).

This opens up avenues to further DRY out the code and make things more async and modern.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,9 +8,11 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import axios from 'axios';
 import { server } from './server.js';
-import MBTiles from '@mapbox/mbtiles';
 import { isValidHttpUrl } from './utils.js';
 import { openPMtiles, getPMtilesInfo } from './pmtiles_adapter.js';
+import { program } from 'commander';
+import { existsP } from './promises.js';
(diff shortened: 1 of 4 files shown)


akx:
Okay, rebased post #1320 merge.

akx:
ping @acalcutt :)

acalcutt:
Overall I do think this looks cleaner and I am likely going to merge this. 

This one thing I am wondering about id the file name of the wrapper. It seems like pmtiles_adapter.js and mbtiles_wrapper.js should be named more similarly. That is just really a nitpick thing.

Speaking of pmtiles, I probably did not promisify it very well, so I am wondering if you think there are any improvements there with pmtiles like these changes.

akx:
> This one thing I am wondering about id the file name of the wrapper. It seems like pmtiles_adapter.js and mbtiles_wrapper.js should be named more similarly. That is just really a nitpick thing.

Maybe. I just named the file after the class name, and the class really is just a simple wrapper around `MBTiles` :)

acalcutt (approved):
Looks cleaner to me. Tested it locally and it seems to work fine

In the future we might want to look at renaming a few of these files, but we can think about it more.

Issue #1324, opened 2024-07-19
Broken rendering - No handling of local GeoJSON files


okimiko:
I'm using the latest docker release (v4.12.0) and created a custom-style based on osm-liberty, which works fine so far in vector and raster view.

Then I integrated a geojson source and added style information for its content. In vector view everything is fine and it works like expected, but in raster view there is no image (grey-screen). The preview on the index page is emtpy, too.

Here the logs, there is no http return code, no size, no duration in error case but no other error:
```
#broken
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty-geojson/512/1/0/0.png - - - - ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty-geojson/512/1/1/1.png - - - - ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty-geojson/512/1/0/1.png - - - - ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty-geojson/512/1/1/0.png - - - - ms

#working
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty/512/1/0/1.png 200 214435 - 259.911 ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty/512/1/0/0.png 200 372680 - 277.889 ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty/512/1/1/1.png 200 211411 - 301.986 ms
mapserver_service.1.sb87pa3vjyv0@devel-08    | GET /styles/osm-liberty/512/1/1/0.png 200 399413 - 343.537 ms
```

And here the snippet of the style:
```
  "sources": {
    "openmaptiles": {
      "type": "vector",
      "url": "mbtiles://{v3}",
      "maxzoom": 7
    },
    "custom-geojson": {
      "type": "geojson",
      "data": "/files/custom.geojson"
    }
  },
[...]
```

I started a bit try and error:
- In vector view it works because my reverse proxy resolves the path
- In raster view the file does not exist because there is no relative parent
   - I tried to add a volume to my docker container at /files with fingers crossed that the file would be found on the local storage
   - It did not work :(

It did not work, too, so I started a bit code digging (serve_rendered.js):
[...]

PR #1326, opened 2024-07-20
Fix  Broken rendering - No handling of local GeoJSON files #1324 

okimiko:
Add general missing handling of local geojson files during rendering (https://github.com/maptiler/tileserver-gl/issues/1324).


acalcutt:
I'm not sure I would want to take this approach. I kind of liked you other idea of adding a "file://" protocol, where you could set up path for "file" paths like we do for others paths https://tileserver.readthedocs.io/en/latest/config.html#paths

For that to work we would need a path set up similar to 
https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L94
and then we would set up serve_rendered.js to handle the file:// protocol.
In your style config you would then prefix your geojson files with file://, similar to how mbtiles:// or pmtiles://

okimiko:
Yes, that behaviour would have been a bit hacky, because it would require a matching environment to work in rendered and vector mode.

I extended the code a bit to support a file:// urls, but stumpled over the spec a bit: geojson uses "data" instead of "url" because it seems to be inteded, to write inline geojson.

In the current implementation it may be a  change with a "bigger" impact, because with /files/ the public interface is extended, but this works without any need of a reverse proxy in both modes.

acalcutt:
I like how this looks better now and will try to test it soon.

One thing I was thinking about was an issue we have with the 'icons' path we added. Right  now, if you don't specify a path for that it defaults to the tileserver-gl root directory, and in that case allows the user to try and load files in the root directory that are not images and causes a crash. In my own instance I specifically set it to the folder `public/resources/images` which avoids that issue, but really it should be hardcoded to a path like that by default, not the root folder like we use now.

I was wondering if we should do the same thing with 'files' to prevent inadvertently exposing all source files and maybe make a default folder specifically for that like `public/resources/files` and use that as the default path if one doesn't exists.  If it was specified in the config it would use the alternative path given, but not set it would default to `public/resources/files`.

acalcutt:
For the comment on geojson sources, it seems to accect inline code or a url
https://maplibre.org/maplibre-style-spec/sources/#geojson

In my own project I have used dynatically generated geojson sources by url on pages like
https://wifidb.net/wifidb/opt/map.php?func=user_list&labeled=0&id=65596&sig_label=none 
which uses url sources

src/serve_rendered.js
```diff
@@ -1020,8 +1020,23 @@ export const serve_rendered = {
                 const format = extensionToFormat[extension] || '';
                 createEmptyResponse(format, '', callback);
               }
+            } else if (protocol === 'file') {
+              const file = url.fileURLToPath(req.url);
```
acalcutt, line 1024: This didn't work for me testing locally on windows

Locally I changed to
```
              const dir = options.paths[protocol];
              const name = unescape(req.url).substring(protocol.length + 3);
              const file = path.join(dir, name)
```
which did seem to work

acalcutt:
I am testing locally with this config

[test_data_geojson.zip](https://github.com/user-attachments/files/16355352/test_data_geojson.zip)

vector map
image

raster map
image


Note, I did make a few changes locally. I changed 'file://' to 'files://' and updated anywhere in the code that just used 'file' to 'files'. This was for consistency with the other protocol and to get it working. I also made the other path change I commented above.

acalcutt:
One thing to note is depending on your geojson file, this is likely going to affect performance because every tile it is basically loading the full geojson file. Were if instead you did like I mentioned before and converted your geojson to mbtiles/pmtiles vector files with [tippecanoe](https://github.com/felt/tippecanoe), it would only need to download the region you are looking at and would be faster.

src/serve_rendered.js
```diff
@@ -1020,8 +1020,23 @@ export const serve_rendered = {
                 const format = extensionToFormat[extension] || '';
                 createEmptyResponse(format, '', callback);
               }
+            } else if (protocol === 'file') {
```
acalcutt, line 1023: change 'file' to 'files' for consitency. this will require you use files:// in your style

src/serve_style.js
```diff
@@ -128,6 +129,14 @@ export const serve_style = {
...
         source.url = `local://data/${identifier}.json`;
       }
+
+      let data = source.data;
+      if (
+        data &&
+        (data.startsWith('file://'))
+      ) {
+          source.data = 'local://files' + data.replace('file://', '').replace(options.paths.files, '');
+      }
```
acalcutt, line 139: Locally i had to change this to 
```
      if (
        data &&
        (data.startsWith('files://'))
      ) {
          source.data = 'local://files/' + data.replace('files://', '').replace(options.paths.files, '');
      }
```

docs/config.rst
```diff
@@ -17,7 +17,8 @@ Example:
         "icons": "icons",
         "styles": "styles",
         "mbtiles": "data",
-        "pmtiles": "data"
+        "pmtiles": "data",
+        "files": "data"
```
acalcutt, line 21: This should proably be 
`"files": "files"` 
just so it doesn't default to sharing your full data directory.

src/main.js
```diff
@@ -115,6 +115,7 @@ const startWithInputFile = async (inputFile) => {
         styles: 'styles',
         mbtiles: inputFilePath,
         pmtiles: inputFilePath,
+        files: inputFilePath
```
acalcutt, line 118: I would just remove this line. It only gets used if a single mbtiles/pmtiles is specfied and in that case, it doesn't really need this files config

src/serve_rendered.js
okimiko: I'm not sure about "files://" vs. "file://". The later is a commonly valid URI protocol (like "http://"), the other one a real custom protocol.

src/serve_style.js
okimiko: See above.

src/serve_rendered.js
okimiko: I will apply that change later, sorry, I have no Windows available for testing.

docs/config.rst
okimiko: For data exposure I had the same thoughts after committing, for my default setup I use "/files", but I think a default beneath "public" will be a good indication, that these files will be exposed!

okimiko:
> One thing to note is depending on your geojson file, this is likely going to affect performance in raster tiles because every tile it makes is basically loading the full geojson file. Were if instead you did like I mentioned before and converted your geojson to mbtiles/pmtiles vector files with [tippecanoe](https://github.com/felt/tippecanoe), it would only need to download the region you are looking at and would be faster, since tippcannoe has already broken your data down into tiles.

Good point! For my current files it seems not to be a problem, but that should be added as note/kept in mind.

I will apply some of the above changes later.

src/serve_rendered.js
acalcutt: This is what the error looks like with the default code. *note* I added a console.log after `const file = url.fileURLToPath(req.url);` just so I could see the path, so you can see the file path is shown in the console multiple times `\\file_65596_2024-07-22_013451.geojson\`
image

acalcutt: my suggested code ends up with a path like this. The top path is 'dir' and the second path is 'file'
image

okimiko: I applied your change and dropped my idea of using "file://" as real uri, because it would allow to expose the whole filesystem.

docs/config.rst
okimiko: Changed to "public/files" per default.

src/main.js
okimiko: In fact I'm using exactly this setup for testing (using the docker image with a slightly modified basic-preview style + added geojson)  ;-)

src/serve_rendered.js
```diff
@@ -1020,8 +1020,23 @@
...
+              if (fs.existsSync(file)) {
+
+                const inputFileStats = fs.statSync(file);
+                if (!inputFileStats.isFile() || inputFileStats.size === 0) {
+                  throw Error(`File is not valid: "${req.url}"`);
+                }
+
+                fs.readFile(file, (err, data) => {
+                  callback(err, { data: data });
+                });
```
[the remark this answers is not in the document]
okimiko: I'm not sure about this, for sprites the same function is used (line 879).

okimiko:
- Merged latest changes (and applied changes due to #1320).
- Tested (and fixed) handling of inline geojson

Dockerfile
```diff
@@ -107,8 +107,8 @@ COPY --from=builder /usr/src/app /usr/src/app
 
 COPY . /usr/src/app
 
-RUN mkdir -p /data && chown node:node /data
-VOLUME /data
+RUN mkdir -p /data /usr/src/app/public/files && chown node:node /data /usr/src/app/public/files
```
acalcutt, line 110: the only thing I don't really like is making this folder with the dockerfile . since tileserver-gl can run directly without docker, it would need that folder there in that use case also. So maybe we should either make the directory with a default file inside, of maybe find a way to create it with the package.json npm prepare..

okimiko: I added the folder to the repository (and excluded .gitignore files for docker build).

acalcutt: sounds goo to me. this can be removed now right?

acalcutt: I meant can we remove the changes in this file. Since the folder exists we don't need to create it here anymore.

src/server.js
```diff
@@ -95,23 +95,18 @@ function start(opts) {
   paths.mbtiles = path.resolve(paths.root, paths.mbtiles || '');
   paths.pmtiles = path.resolve(paths.root, paths.pmtiles || '');
   paths.icons = path.resolve(paths.root, paths.icons || '');
+  paths.files = path.resolve(paths.root, paths.files || 'public/files');
```
acalcutt, line 98: Nice, so if the paths root an files don't exists, the deafult publick files get used?

Can you do me a favor and fix a bug? Can you do the same thing to the icons path right above this and set the default to 'public/resources/images' .

Dockerfile
okimiko: Sorry, I was a bit confused.

In fact the `mkdir` for public/files may be removed, but due to the '-p' it does not hurt and ensures, that the path, which will be chown'ed and exposed as volume is really there. 
If I remove/change it, I would revert the whole file, because the volume depends on the configuration and everything will still  work with the default configuration, because the mount may be applied without a exposed volume of an empty directory. (This may be different, if there was a default config file included in the image.)

In short: It's up to you ;-)

src/server.js
okimiko: If the config object icons does not exist, yes. The path existence check itself follows in the next code block.

I added the change for icons.

Dockerfile
acalcutt: I would rather keep it as it was before since it shouldn't really be needed no.

src/server.js
acalcutt: thanks!

Dockerfile
okimiko: Done

okimiko:
@acalcutt I merged the latest changes, applied the prettier fixes and finally updated the tests to be executed in the main directory, so no further stubbing should be required.

acalcutt (approved):

okimiko:
@acalcutt: I tidied up a little bit (merge master, lint:*:fix, squash - https://github.com/okimiko/tileserver-gl/commit/222d6c15e76a07b3ac63e43c7644f6b7171aa5f9). If you want to merge, this may be a bit cleaner than this here. If wanted, I can force push it or we close this here and I create a new, clean PR.

src/serve_rendered.js
```diff
@@ -893,13 +893,13 @@ export const serve_rendered = {
             // console.log('Handling request:', req);
             if (protocol === 'sprites') {
               const dir = options.paths[protocol];
-              const file = unescape(req.url).substring(protocol.length + 3);
+              const file = decodeURI(req.url).substring(protocol.length + 3).replaceAll('%2c', ',');
```
acalcutt, line 896: What is this replacement added for?

okimiko: Sorry, I did not check the code before. Correct would be to use `decodeURIComponent` as counterpart of `encodeURIComponent`.
`decodeURI` / `encodeURI` does not escape commas, so I added the extra replace after using a style with two fonts configured and receiving `ERROR: Font not found: Noto Sans Regular%2cNoto Sans Bold`.

I will fix that later today an force push the squashed commit.

PR #1327, opened 2024-07-22
Replace `formatQuality` with `formatOptions` for additional configurability

nathreed:
Per the (brief) discussion https://github.com/maptiler/tileserver-gl/discussions/1281, this PR replaces the `formatQuality` option in the config file with `formatOptions`. 

The documentation is also updated with examples for the previous JPEG and WebP quality settings as well as the new PNG options.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -504,6 +504,15 @@ const respondImage = (
         image.composite(composites);
       }
 
+      // Legacy formatQuality is deprecated but still works
+      const formatQualities = options.formatQuality || {};
+      if (Object.keys(formatQualities).length !== 0) {
+        console.log(
+          'WARNING: The formatQuality option is deprecated and has been replaced with formatOptions. Please see the documentation. The values from formatQuality will be used if a quality setting is not provided via formatOptions.',
+        );
+      }
+      const formatQuality = formatQualities[format];
+
       const formatOptions = (options.formatOptions || {})[format] || {};
 
       if (format === 'png') {
@@ -518,9 +527,9 @@ const respondImage = (
           dither: formatOptions.dither,
         });
       } else if (format === 'jpeg') {
-        image.jpeg({ quality: formatOptions.quality || 80 });
+        image.jpeg({ quality: formatOptions.quality || formatQuality || 80 });
       } else if (format === 'webp') {
-        image.webp({ quality: formatOptions.quality || 90 });
+        image.webp({ quality: formatOptions.quality || formatQuality || 90 });
       }
       image.toBuffer((err, buffer, info) => {
         if (!buffer) {
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -504,14 +504,23 @@ const respondImage = (
         image.composite(composites);
       }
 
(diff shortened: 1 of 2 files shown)


src/serve_rendered.js
```diff
@@ -503,14 +503,23 @@ const respondImage = (
         image.composite(composites);
       }
 
-      const formatQuality = (options.formatQuality || {})[format];
+      const formatOptions = (options.formatOptions || {})[format] || {};
```
akx, line 506: Feels like `options.formatQuality` should still be supported (IOW, when people would update to a new version, they suddenly wouldn't get lower or higher qualities than what they expect), maybe log a deprecation warning though?

nathreed: ok, this is done. `formatQuality` is still supported but will produce a deprecation warning if it's found. `formatOptions` is preferred over `formatQuality`.

acalcutt (approved):
I think this looks good and offers a good path for people using the old option.

One thing to note is when testing the 4 colors noted in the documentation I did see some text that had different colors, i think it was  text that crosses the tile boarder.
image

not sure this is fixable since the colors are being determined per tile

nathreed:
> One thing to note is when testing the 4 colors noted in the documentation I did see some text that had different colors, i think it was text that crosses the tile border.

yeah, I noticed this in testing too. Unfortunately `sharp` doesn’t seem to let you specify _which_ four colors you want. Maybe we should pick a different example for the documentation, one that people would be more likely to use? I think it’s fine as-is (people would probably find this pretty quick) but happy to change it if you think we should.

acalcutt:
I wouldn't mind a PR to change the docs to better defaults that are less likely to cause issues.

ampledata:
Should this configuration parameter work with mbtiles with png tiles? For example, if I use mapTiler to build a mbtiles file, and select PNG as the output format, can I assume that `formatOptions` will apply to those PNGs? I ask because I've tried this but haven't noticed any options being applied. I'm testing on the v5.0.0 Docker image.

acalcutt:
It would not apply to your data files, so it would not change the png images inside your mbtiles.

tiles that are rendered like the style raster endpoints or static image endpoints would be changed by this.

ampledata:
I understand this will not change the files within the mbtiles, and to clarify I meant the rendered tiles that the tileserver serves from within the mbtiles - that is, will this change apply to the raster tiles rendered on the fly FROM the mbtiles file, served by tileserver? I guess I don't totally understand how these options are apply, or where, or to what.

acalcutt:
It will not change the pngs served from you data files either. It serves those as is

Like i said, it only affects rendered images, so that would be the rendered 'raster' endpoint under styles or the image endpoints under styles. 
https://tileserver.readthedocs.io/en/latest/endpoints.html#rendered-tiles
https://tileserver.readthedocs.io/en/latest/endpoints.html#static-images
https://tileserver.readthedocs.io/en/latest/endpoints.html#wmts-capabilities


If you wanted to render your data files with different format options, you could make a style that uses that data source and inclue it as a layer. then the resulting images from your style would be rendered with those options formatOptions

PR #1335, opened 2024-07-29
feat: simple typeahead search

mvexel:
This adds a simple typeahead search to the layer and style overview. Useful for instances with a lot of layers!

https://github.com/user-attachments/assets/e230b8f9-0c2e-44b7-a095-f2b33f501bb7


public/templates/index.tmpl
```diff
@@ -13,12 +13,42 @@
...
+          continue;
+        }
+        var details = item.getElementsByClassName('details')[0];
+        var name = details.getElementsByTagName('h3')[0].textContent.toLowerCase();
+        var identifier = details.getElementsByClassName('identifier')[0].textContent.toLowerCase().substring(12); // remove "identifier: "
+        if (name.indexOf(filter) > -1 || identifier.indexOf(filter) > -1) {
+          item.style.display = 'block';
+        } else {
+          item.style.display = 'none';
+        }
```
akx, line 32:
```suggestion
        item.hidden = !(name.indexOf(filter) > -1 || identifier.indexOf(filter) > -1);
```

public/templates/index.tmpl
```diff
@@ -13,12 +13,42 @@
...
+
+    function filter() {
+      var filter = document.getElementById('filter').value.toLowerCase();
+      var items = document.getElementsByClassName('item');
+      for (var i = 0; i < items.length; i++) {
+        var item = items[i];
+        if (!item.getElementsByClassName('details')[0]) {
+          continue;
+        }
+        var details = item.getElementsByClassName('details')[0];
```
akx, line 25:
```suggestion
        var details = item.getElementsByClassName('details')[0];
        if(!details) continue;
```

public/templates/index.tmpl
```diff
@@ -13,12 +13,42 @@
...
+        var name = details.getElementsByTagName('h3')[0].textContent.toLowerCase();
+        var identifier = details.getElementsByClassName('identifier')[0].textContent.toLowerCase().substring(12); // remove "identifier: "
```
akx, line 27:
```suggestion
        var identifier = details.getElementsByClassName('identifier')[0].textContent.toLowerCase().replace(/^identifier: /, '');
```

akx (commented):
I'd suggest adding [custom data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) to the elements being filtered, so we don't need to traverse the DOM as much :)

mvexel:
@akx haha yes that would be better for sure!
Do you otherwise see this as a useful thing to merge (before I spend more time on it)?

acalcutt:
I think this looks like a nice feature i would merge.

mvexel:
Thanks @acalcutt! I will make the improvements @akx suggested to reduce strain from DOM traversal.

mvexel:
all done and as promised we are *actually* filtering on both name and identifier now :) 
We could extend this with another data attribute to filter on style, layer only with some checkboxes but I propose not overcomplicate things.
Thanks for considering!

akx:
Looks nice and clean with `hidden` and the dataset stuff! Thanks for taking my suggestions into account :)

mvexel:
> Looks nice and clean with `hidden` and the dataset stuff! Thanks for taking my suggestions into account :)

Thanks for the suggestions! I hope this is ready to merge but please do let me know if you have any additional changes you would like to see.

acalcutt:
Just an FYI, I will try to test this soon so it can be merged. I have just been slammed this week at my real job and haven't had much time to test PRs this week.

acalcutt (approved):
Seems simple and worked fine in testing.

acalcutt:
The one thing I haven't liked about this is how much room the logo + filter take up. one of these days I need to shrink that logo

mvexel:
Have to agree with you. With a little bit of work the filter element could be made smaller. Honestly I think the entire page could benefit from a tighter layout. Let me think about this a bit.

PR #1397, opened 2024-09-26
feat: add progressive rendering option for static jpeg images

samuel-git:
Adds support for progessive rendering of static jpeg images via the ``formatOptions``.
In my testing, progressive jpegs were ~2% smaller in filesize than their non-progressive counterparts.
They also 'feel' faster because they become more sharp while loading instead of loading in rows from top to bottom.

Also updated the section in the docs.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -94,7 +94,8 @@ Use ``false`` to disable the front page altogether (404).
 -----------------
 
 You can use this to specify options for the generation of images in the supported file formats.
-For JPEG and WebP, the only supported option is ``quality`` [0-100].
+For WebP, the only supported option is ``quality`` [0-100].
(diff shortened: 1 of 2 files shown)


acalcutt:
seems like a good addition. can you run 'npm run lint:js:fix' on this to fix the ci lint error

samuel-git:
> seems like a good addition. can you run 'npm run lint:js:fix' on this to fix the ci lint error

Thanks @acalcutt for the heads up, updated the commit.

acalcutt:
The only other commend I have is it would be nice if the readme included an emable of what it want for boolean. Does it want [0 or 1] or [true or false]

samuel-git:
Thanks, updated!

acalcutt (approved):

Issue #1412, opened 2024-11-08
Security Vulnerabilities - Request to update express version from 4.19.2 to 4.21.1


tblauer:
When running anchor and trivy vulnerability scans on this library, there are 6 CVEs showing up that are all associated with the version of express and/or it's dependencies.

Updating to express 4.21.1 would update all of the affected libraries to versions in which the vulnerabilities have been fixed

The table below shows the affected libraries, the fixed version according to the CVE and which version of the dependent libraries are in the specified version of express

| CVE                       | Package             | Installed Version   | Fixed In             | express 4.20.0  | express 4.21.0  | express 4.21.1 |
| :------------------  | :----------------  | :------------------  | :----------------- | :---------------  | :--------------- | :--------------- |
| CVE-2024-43796  | express              | 4.19.2                    | 4.20.0                 | 4.20.0               | 4.21.0              | 4.21.1              |
| CVE-2024-45296  | path-to-regexp | 0.1.7                      | 0.1.10 or 8.0.0    | 0.1.10               | 0.1.10               | 0.1.10              |
| CVE-2024-45590  | body-parser      | 1.20.2                    | 1.20.3                 | 1.20.3               | 1.20.3               | 1.20.3              |
| CVE-2024-43799  | send                  | 0.18.0                    | 0.19.0                 | 0.19.0               | 0.19.0               | 0.19.0              |
| CVE-2024-43800  | serve-static       | 1.16.0                    | 1.16.0                 | 1.16.0               | 1.16.2               | 1.16.2              |
| CVE-2024-47764  | cookie               | 0.6.0                      | 0.7.0                  | 0.6.0                 | 0.6.0                 | 0.7.1                 |

acalcutt:
Seems like a failed test stopped that from getting updated in https://github.com/maptiler/tileserver-gl/pull/1401 . If someone wants to look into that we can move it forward

acalcutt:
If you have a chance could you evaluate this PR to upgrade express to v5 and let me know if it has any issues for you
https://github.com/maptiler/tileserver-gl/pull/1429

acalcutt:
Should be Fixed by https://github.com/maptiler/tileserver-gl/pull/1429 , which is included as of 5.1.1 (actually 5.1.0, but there was a docker issue in that version)

PR #1424, opened 2024-12-17
Fix failing test due to updated path-to-regexp

okimiko:
Workaround was found here: https://github.com/pillarjs/path-to-regexp/security/advisories/GHSA-9wv6-86v2-598j

Should fix/support #1412.

### DIFF
--- a/src/serve_rendered.js
+++ b/src/serve_rendered.js
@@ -855,7 +855,7 @@ export const serve_rendered = {
       );
     }
 
-    app.get('/(:tileSize(256|512)/)?:id.json', (req, res, next) => {
+    app.get('/(:tileSize(256|512)/)?:id([^/]+).json', (req, res, next) => {
       const item = repo[req.params.id];
       if (!item) {
         return res.sendStatus(404);
--- a/test/static.js
+++ b/test/static.js
@@ -123,10 +123,11 @@ describe('Static endpoints', function () {
       });
 
       describe('different parameters', function () {
-        testStatic(prefix, '-180,-90,180,90/20x20', 'png', 200, 2);
+        //invalid center
+        testStatic(prefix, '-180,-90,180,90/20x20', 'png', 400, 2);
         testStatic(prefix, '0,0,1,1/200x200', 'png', 200, 3);
-
-        testStatic(prefix, '-280,-80,0,80/280x160', 'png', 200);
+        //invalid center
+        testStatic(prefix, '-280,-80,0,80/280x160', 'png', 400);
       });
     });
 
--- a/test/static.js
+++ b/test/static.js
@@ -123,15 +123,16 @@ describe('Static endpoints', function () {
       });
 
       describe('different parameters', function () {
-        //invalid center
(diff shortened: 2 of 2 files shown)


acalcutt:
The ci tests didn't seem to run in this PR, so i tested merging this into https://github.com/maptiler/tileserver-gl/pull/1401 . however it doesn't seem to fix the issue

I tested this locally and I get
Invalid regular expression: /^/(?:(?:(256|512))/)?(?:((?:(?!/|//)?).)+?))\.json/?$/i: Unmatched ')'

okimiko:
I can not confirm this: I did a clean checkout and build a clean docker build-image (based on the build step in the default Dockerfile) and did not have this error in my environment.

But in fact, the runtime tests did not work because the process failed to spawn (missing X display), so I only tested my change, which worked so far. In https://github.com/maptiler/tileserver-gl/actions/runs/11783668480/job/32821321983#step:10:169 we can see the same, but there is a different path, which needs an update, too.

I just added the missing `xvfb-run` to `npm test` and can confirm the results of #1401. I will try to take care of that later.

okimiko:
I found no issues in the routes, I even reviewed the express routes with a route tester (fyi: https://forbeslindesay.github.io/express-route-tester/). But at last I had the idea to look at the result of the test and received a "invalid center", which is correct (see https://github.com/maptiler/tileserver-gl/blame/master/src/serve_rendered.js#L384-L390).

So I fixed the test results for the related queries, moved them to the "invalid requests" section and added correct ones in "different parameters".

It seems, that with the previous version of express/path-to-regexp the test passes (and a running 5.0.0 instance is in fact delivering an image), so the path handling seems to be mixed up or at least `respondImage`  does not receive the out-of-range parameters.

acalcutt:
After more testing I see this does fix this error
image

With your fix the tests get past that issue, and actually run. However like you mention, tjere are still two test failing that are preventing the upgrade of express.
image
(this is run with the steps in 
image  ...exceot on windows i don't need xvfb-run )

Like you mentioned, I noticed that 'Invalid Center' also. I did test older versions and can confirm it worked when using the older express. I did spend some time looking at the regex's . I think that may be handled by https://github.com/maptiler/tileserver-gl/blob/master/src/serve_rendered.js#L751 , but it is a bit confusing. I had gotten a few suggestions from google AI but didn't have any luck with them. It would seem to get 'Invalid Center'  it must not be getting the coordinates

okimiko:
Yes, `serveBounds` is correct and should lead to `respondImage` and finally to the out-of-bounds check -> "Invalid Center".

I think the issue was already in the code and just revealed by the update. You may merge your branch again, I did two fixed in the test config:
- Set correct return code for current tests
- Add new ones, with valid parameters for this case

I do not realy understand, why the test passed before, but I assume that something has changed in the environment or a different routing expression was used in the past (because of a weaker regex?). Anyway, this issue helped me finding the issue in #1425 ;-)

okimiko:
Obsolete due to #1429.

PR #1427, opened 2024-12-27
Use npm packages for public/resources

okimiko:
I wanted to add a dependency to my preview page and stumbled over the mirrored files and your discussion #1208

Here is my suggestion as "prepare" script including (the latest versions of)
- leaflet
- leaflet-hash
- maplibre-gl
- maplibre-gl-inspect

an updated Dockerfile (in general -> Using "&&" instead of ";" to chain commands) and especially with the added "copyfiles" regarding the new "prepare" step.


acalcutt:
The only thing that concerns me with this is dependabot automatically upgrade to major releases. I think we need to double check the dependabot settings because it automatically merged the last major maplibre-native and I didn't expect it would.

I know the next release of leaflet is a breaking change because the L. syntax is going away, although I am pretty sure the code we have that uses it doesn't actually do anything (since it deals with pixel ration, which leaflet needs @2x in the (url) path...not like our preview page uses.

and i know maplibre just experienced so pushbash with their geometry type change that ended up getting reveret

point being, i think this would be fine as long as we make sure dependabot is fixed so it doesn't automerge major releases.... but I still need to look at that.

okimiko:
Maybe this helps (untested, just of the docs):
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#ignore -> `version-update:semver-major`

```
version: 2
updates:
  - package-ecosystem: npm
    versioning-strategy: increase
    directory: '/'
    schedule:
      interval: daily
    ignore:
      - dependency-name: "leaflet"
        update-types: ["version-update:semver-major"]
      - dependency-name: "leaflet-hash"
        update-types: ["version-update:semver-major"]
      - dependency-name: "@mapbox/mapbox-gl-rtl-text"
        update-types: ["version-update:semver-major"]
      - dependency-name: "@maplibre/maplibre-gl-inspect"
        update-types: ["version-update:semver-major"]
      - dependency-name: "maplibre-gl"
        update-types: ["version-update:semver-major"]
    commit-message:
      prefix: fix
      prefix-development: chore
      include: scope
  - package-ecosystem: github-actions
    directory: '/'
    schedule:
      interval: weekly
    commit-message:
      prefix: fix
      prefix-development: chore
      include: scope
```

acalcutt:
based on this it looks like we could set the target to minor
https://github.com/fastify/github-action-merge-dependabot?tab=readme-ov-file#specifying-target-for-development-and-production-packages-separately

okimiko:
Check. For automerger exists a `target` parameter, too:

> A flag to only auto-merge updates based on Semantic Versioning.
Possible options are: major, premajor, minor, preminor, patch, prepatch, prerelease, any.

okimiko:
From my perspective it does not make sense to handle development different than production: Major releases may cause problems in both cases. But feel free to change it, if you have a different opinion or experience.

acalcutt:
I agree that we should just target minor. and and major updates can be manually approved

I did just update the release workflow to allow pre-releases,, so in the future i think it will a low some level of testing. in the past this project didn't have pre-releases.

acalcutt (approved):

Issue #1411, opened 2024-11-06
Random crashes when rasterizing tiles


georgbachmann:
We are using tileserver-gl to serve our maps. We have apps that only use vector tiles and that is NO PROBLEM at all. We separated all incoming requests already and noticed that serving pure vector data does not cause any problems. But for our website we still use raster tiles and that seems to cause problems after a couple of hours. We get consistent crashes and we think (not know) that it has something todo with memory consumption?!? At least memory seems to spike before crashes happen. The machines would have plenty though (50GB) but it's already crashing way before that.
Currently we use version 4.11.1 and also tried the latest 5.0.0 to test and still the same problem. It's running on linux machines in a docker-container. (`maptiler/tileserver-gl:v4.11.1`)

We can also not give more information about what causes the issue, cause we don't know it and the logs of our docker-container don't say much.

So my first question would be: how to provide information to narrow down the problem?

acalcutt:
Does this seem similar to https://github.com/maptiler/tileserver-gl/issues/1236 . Possibly try 4.5.1 and see if the issue happens in that version.

In docker are you able to see how much memory is allocated and in use? My understanding is y default all memory is available, but maybe that isn't the case?

I run 5.0.0 without the docker here on Ubuntu 22.04 and haven't noticed any issues, but some people have reported similar issues when using the docker.

georgbachmann:
@acalcutt thank you very much for helping!!!
We don't think this issue is related - as far as we can tell, none of our systems ever had zombie processes and those seemingly random crashes have been an ongoing issue for us for years now: Having tried dozens of versions all over 4.x (I think even 3.x) up to 5.0.0 now, doesn't seem to change much if anything at all.
[...]

PR #1429, opened 2024-12-29
Upgrade Express to v5 +Canvas v3 + code cleanup

acalcutt:
This PR addresses https://github.com/maptiler/tileserver-gl/issues/1412 by upgrade express to v5. It also tries to address https://github.com/maptiler/tileserver-gl/issues/1411 by adding Try/Catch around that function and also adding better verbose logging. It also upgrades canvas to address https://github.com/maptiler/tileserver-gl/issues/1433

express v5 features a new version of 'path-to-regexp', which changes the syntax to no longer allow optional parameters by surrounding them with ( and )? and  putting the type inside ( and ). see https://github.com/pillarjs/path-to-regexp?tab=readme-ov-file#errors
```
Unexpected ? or +
In past releases, ?, *, and + were used to denote optional or repeating parameters. As an alternative, try these:
For optional (?), use an empty segment in a group such as /:file{.:ext}.
For repeating (+), only wildcard matching is supported, such as /*path.
For optional repeating (*), use a group and a wildcard parameter such as /files{/*path}.
Unexpected (, ), [, ], etc.
Previous versions of Path-to-RegExp used these for RegExp features. This version no longer supports them so they've been reserved to avoid ambiguity. To use these characters literally, escape them with a backslash, e.g. "\\(".
```

In tileserver-gl we had a lot of spots we were using optional parameters and defining types by regex values, which because of this change of  'path-to-regexp' used by express, are no longer supported

This PR tries to address this change by doing the following
1.) All optional parameters have been changed from the format (/:spriteID)?, to the new format {/:spriteID}
2.) All validation has been removed from the enpoints, For example, :range([\\d]+-[\\d]+) became :range
3.) Because validation was removed from the endpoints, some of our paths under /styles/ became too similar. I had to merge these into a single path and add in some logic to direct the request the right way.
4.) Also because validation was removed I tried to add in some logic for accepted values for things like tileSize and Scale
5.) The syntax we used in ServeTemplate also has to change. the '$' was no longer supported in the path, so it had to be removed. This lead to a need to use next('route') to redirect request not served by the template.

Other changes. 
1.) Tried to sanitize outputs. I  still wanted some information in the logs, so it tries to at least sanitize the data to acceptable characters. codeql seems to still not like this.
2.) Tried to change variable style functions to traditional functions for better logging information
3.) Added generated function headers so the code is more informative.
4.) Some status codes have change. for example 400 vs 404 when bad data is provided (test had to be updated for this). Also, when a tile is not found it now returns a [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) to try and allow sparse tile suport.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -48,11 +48,11 @@ export const serve_data = {
       if (verbose) {
         console.log(
           `Handling tile request for: /data/%s/%s/%s/%s.%s`,
-          String(id).replace(/\n|\r/g, ''),
-          String(z).replace(/\n|\r/g, ''),
-          String(x).replace(/\n|\r/g, ''),
-          String(y).replace(/\n|\r/g, ''),
-          String(format).replace(/\n|\r/g, ''),
+          String(req.params.id).replace(/\n|\r/g, ''),
+          String(req.params.z).replace(/\n|\r/g, ''),
+          String(req.params.x).replace(/\n|\r/g, ''),
+          String(req.params.y).replace(/\n|\r/g, ''),
+          String(req.params.format).replace(/\n|\r/g, ''),
         );
       }
       const item = repo[req.params.id];
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -165,7 +165,11 @@ export const serve_data = {
       try {
         if (verbose) {
           console.log(
-            `Handling elevation request for: /data/${req.params.id}/elevation/${req.params.z}/${req.params.x}/${req.params.y}`,
+            `Handling elevation request for: /data/%s/elevation/%s/%s/%s`,
+            String(req.params.id).replace(/\n|\r/g, ''),
+            String(req.params.z).replace(/\n|\r/g, ''),
+            String(req.params.x).replace(/\n|\r/g, ''),
+            String(req.params.y).replace(/\n|\r/g, ''),
           );
         }
         const item = repo?.[req.params.id];
@@ -313,7 +317,8 @@ export const serve_data = {
     app.get('/:id.json', (req, res) => {
       if (verbose) {
         console.log(
-          `Handling tilejson request for: /data/${req.params.id}.json`,
+          `Handling tilejson request for: /data/%s.json`,
+          String(req.params.id).replace(/\n|\r/g, ''),
         );
       }
       const item = repo[req.params.id];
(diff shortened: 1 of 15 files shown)


okimiko:
I just had a short look at the code changes; as far as I have seen it looks OK so far.

I build the image and used it with my test config/data and can see a lot of "Handling request" messages on the application console and for some styles I have errors regarding fonts in the browser console. I would like to take a deeper look at that and test verbose logging a bit. May be I can dig out some older scripts, which use the static endpoint, too.

In addition I would like to apply my elevation fix to this PR and test again. I hope I will have time in the next few days.

acalcutt:
Do you have verbose enabled in your test config? You should only see the handling messages when verbose is enabled. I added more logging in verbose mode to help troubleshoot this.

acalcutt:
One concert I have is with fonts. I added a lot of validation there to make codeql happy, but i'm not sure they are all good changes.

okimiko:
I re-checked everything:
- I start the tileserver with "--verbose", without the logs are fine.
- The font errors I've seen are related to open sans, which I use in my setup (https://bugzilla.mozilla.org/show_bug.cgi?id=1185685)
- I started my map-exporter script, which uses the static endpoint and everything looks fine so far.

src/serve_data.js
```diff
@@ -11,338 +11,316 @@ import SphericalMercator from '@mapbox/sphericalmercator';
...
 
-          const TILE_SIZE = 256;
-          let tileCenter;
-          let xy;
-
-          if (Number.isInteger(x) && Number.isInteger(y)) {
-            const intX = parseInt(req.params.x, 10);
-            const intY = parseInt(req.params.y, 10);
-
+        const fetchTile = await fetchTileData(source, sourceType, z, x, y);
```
okimiko, line 240: Use `const fetchTile = await fetchTileData(source, sourceType, z, xy[0], xy[1]);` instead. This was invalid for pmtiles before.

okimiko (commented):
I applied and adjusted changes of #1432 in [another branch](https://github.com/okimiko/tileserver-gl/tree/express-5).

Except the coordinates of fetchTileData (which is fixed in the new branch, too) this PR works for me.

okimiko (approved):

okimiko (approved):

okimiko (approved):
Tested elevation again due to the canvas update; still all fine.

Issue #1448, opened 2025-01-15
tileserver-gl-light no longer working as of 5.1.0


drew-altana:
So 5.0.0 tileserver-gl-light works great. When upgrading to 5.1.0 or 5.1.1 it fails on a canvas dependency (which wants pixman and cairo and all the heavy weight dependencies).
 

```
 npm install -g tileserver-gl-light@5.1.0 
npm warn deprecated @npmcli/move-file@1.1.2: This functionality has been moved to @npmcli/fs
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated npmlog@6.0.2: This package is no longer supported.
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated are-we-there-yet@3.0.1: This package is no longer supported.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated advanced-pool@0.3.3: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm warn deprecated gauge@4.0.4: This package is no longer supported.

changed 353 packages in 6s

40 packages are looking for funding
  run `npm fund` for details
root@04348d187df8:~# tileserver-gl-light 
node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'canvas' imported from /opt/nodejs/lib/node_modules/tileserver-gl-light/src/serve_data.js
    at new NodeError (node:internal/errors:405:5)
    at packageResolve (node:internal/modules/esm/resolve:916:9)
    at moduleResolve (node:internal/modules/esm/resolve:973:20)
    at defaultResolve (node:internal/modules/esm/resolve:1206:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:404:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:373:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:250:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:39)
    at link (node:internal/modules/esm/module_job:75:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v18.20.5
```

[...]

PR #1449, opened 2025-01-16
Fix broken light (invalid use of heavy dependencies)

okimiko:
- Move elevation calculation to `serve_rendered` and stub in `serve_light` due to use of `canvas` and `sharp`
- Elevation API is not available for light!

Possible fix for #1448

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -108,6 +108,8 @@ Source data
 
     * the result will be a json object like ``{"z":7,"x":68,"y":45,"red":134,"green":66,"blue":0,"latitude":11.84069,"longitude":46.04798,"elevation":1602}``
 
+  * The elevation api is not available in the ``tileserver-gl-light`` version.
+
 Static files
 ===========
 * Static files are served at ``/files/{filename}``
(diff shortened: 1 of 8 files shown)


acalcutt:
Do you know if the is a isLight flag in the template? it might be goot to not show the elevation display on the terrain page..

acalcutt:
I am fine with moving this to serve_rendered.js for now. though I have been thinking about ways to shorten that file since it is quite large (like separating it out into other files a bit)

maybe we could make a terrain.js file and for now have it access through rendering so we don't need a new dummy file.. idk. this could be a first step though to at least get light working.

okimiko:
Yes, in general it not the best solution, but may be I had chosen this before, too, just to not have multiple locations with different code. And in fact because we use images here, it is a kind of rendering, too.

I tested light and normal again, looks fine now and finaly: I removed the control in light mode, too :)

acalcutt:
can you also hide this elevation url? since that is unavalable in light

image

okimiko:
Done. I added a note to the docs, too.

acalcutt:
Does preview terrain get hidden when that link is gone? It seemed like that part still worked ok

okimiko:
Yes, now the links are handled in dedicated flags.

acalcutt (approved):
tested and in looks good to me

Issue #1383, opened 2024-09-03
Docker: "Startup complete" fails for unknown reason after SIGHUP


larsschwarz:
I'm running the dockered version and update some mbtiles files frequently which requires a SIGHUP to reload the server. 

This works fine for like 2 days, but after like 2 days the server becomes unavailable. The log always looks like this

```
Caught signal SIGHUP, refreshing
Stopping server and reloading config
Starting server
Listening at http://[::]:8080/
Startup complete
Caught signal SIGHUP, refreshing
Stopping server and reloading config
Starting server
Listening at http://[::]:8080/
```
As you can see "startup complete" isn't logged in those cases. 

Afaik there is no more verbose mode available, so I have no idea how to debug this issue. On a side note I was wondering why there are no timestamps printed out. Without them it's even harder to check other logs for issues occured at the same time that might be related.

Happens with Docker 27.1.2, build d01f264 running Ubuntu 22.04.4 LTS.

acalcutt:
Which version on TileServer-GL? Is there an version that was working for you? "Startup complete" seems to come from https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L615  and "Listening at" seems to come from https://github.com/maptiler/tileserver-gl/blob/master/src/server.js#L635 .  

It looks like startup complete get show when all the promises are loaded, so it seems like maybe something isn't completing. not sure what that would be....

larsschwarz:
Version is v4.12.0, haven't tried the latest yet. Haven't tried any previous versions before 4.1.2.0 either. 

Seems like the health check obviously fails afterwards, however Docker does not seem to restart the container even though I tried using `--restart unless-stopped` to circumvent the issue. 

Guess I have to look into adding a few more debug logs in server.js then, unfortunately it runs just fine for approximately two days with a SIGHUP reload every 15 minutes so it's a little bit hard to debug this specific issue 😩

acalcutt:
Yes I hate bugs like that. if you are restarting every 15 minutes you would think you would see that more often... maybe something like memory filling up in the docker image? I had read on issue somewhere that the SIGHUP restart didn't seems like it was releasing memory.

acalcutt:
I would also see if 4.13.3 or 5.0.0 behave differently, since there have been a few bug fixes relating to fonts in 4.12.0 which caused a bit of stangness rendering for me

larsschwarz:
[...]

Issue #585, opened 2022-04-13
Memory leak kubernetes (docker) maptiler/tileserver-gl 


vkarabchuk:
Hi all
I have a problem with memory leak issue when running maptiler/tileserver-gl in kubernetes
I see that the node process used full the RAM until it hits the limit as a result the node reboot
I have 3 replicas on nodes with 4 cpu and 8 ram in hetzner (ubuntu 20.04)
I am using 3.1.1 tilesserver version
from the changes, I added crons to the container and also installed lsof and reread the mbtile file that changes every 30 minutes 
using the kill -HUP $(lsof -t -i:80) command and i add osm-bright styles
6 GB of RAM using all in 6 hours
kubernetes v1.19.1
kubernetes with docker containerization
Could you help to solve the issue?

alexriabtsev:
@petrsloup maybe you could help?

boldtrn:
Does this issue still exist with v4.0.0?

basduineveld:
We are running v4.4.2 and are seeing the same issue. With every HUP signal, the memory usage increases by about 200 MB, until the pod goes OOM. When we disable the cronjob that sends the HUP signal, the memory usage remains stable at 800 MB for days.

acalcutt:
If anyone has a solution I welcome a PR. for me I don't use docker here, i just run tileserver-gl directly in node, but it doesn't seems to have the same memory issues when run that way (at least I have never experienced memory issues with tileserver run directly)

PR #1455, opened 2025-01-31
Fix SIGUP Memory Leak

okimiko:
This addresses #585 and may address #1383.

I stepped down the code and found out, that the memory issue does not apply to the light build. After some debugging I found
https://github.com/maptiler/tileserver-gl/blob/a245126c20400e76d851e269c0381dbdde926b5e/src/serve_rendered.js#L1060
and 
https://github.com/maptiler/tileserver-gl/blob/a245126c20400e76d851e269c0381dbdde926b5e/src/serve_rendered.js#L1220

I assumed that these are not found by the garbage collector because there is a reference somewhere, to I added an explicit cleanup function `clear`, which required the before inaccessible `serving` to handle the `repo`.

As a result I have no more increasing memory with reloading by SIGHUP.

### DIFF
--- a/src/serve_light.js
+++ b/src/serve_light.js
@@ -6,6 +6,7 @@ export const serve_rendered = {
   init: (options, repo, programOpts) => {},
   add: (options, repo, params, id, programOpts, dataResolver) => {},
   remove: (repo, id) => {},
+  clear: (repo) => {},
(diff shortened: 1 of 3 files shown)


acalcutt (approved):
nice work tracking this down.

PR #1459, opened 2025-02-05
Improve support for X-Forwarded-Path

moongladestorm:
X-Forwarded-Path is already supported in e.g. getTileUrls() from utils.js, but missing in other places.

This fixes the urls with X-Forwarded-Path for all templates (data.tmpl, viewer.tmpl, index.tmpl, wmts.tmpl).

This pull request adds support to the public_url from serveTemplate() function, used by data.tmpl, viewer.tmpl and index.tmpl. Additionally it adds support to baseUrl from serving wmts.xml.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -476,7 +476,8 @@ async function start(opts) {
           if (data) {
             data['server_version'] =
               `${packageJson.name} v${packageJson.version}`;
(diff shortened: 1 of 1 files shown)


acalcutt:
Don't the other pages do this through utils.js instead of doing it themselves? I thought here was a function to create the public url. I will look later when i have some time.

PR #1462, opened 2025-02-10
feat: Fetch style from url

### LINKED ISSUE #1460: Feature to fetch style.json from a url
It would be nice to be able to define a stylesheet from a url in the configuration. I imagine it would look something like this in the config:

```json
"styles": {
    "remote": {
      "style": "https://api.maptiler.com/maps/streets-v2/style.json?key=********************",
      "tilejson": {
        "type": "overlay"
      }
    },
}
```

Would this be possible? I was thinking that the style could be fetched when the app is launched (or if the config is reloaded using SIGHUP). I could probably add this feature myself, but would like to get the opinion of any contributors first in case there is a good reason not to add this.

YoelRidgway:
This PR adds the ability to fetch remote JSON stylesheets by specifying a url in the config.

For example:
```json
"styles": {
    "streets-v2": {
      "style": "https://api.maptiler.com/maps/streets-v2/style.json?key=********************",
      "tilejson": {
        "type": "overlay"
      }
    },
    "maplibre-demotiles": {
       "style": "https://demotiles.maplibre.org/style.json"
    }
}
```
Resolves #1460


acalcutt (approved):
I tested this and it seems to work ad expected. looks good to me.

Sorry for the delay in looking at this, I have been busy with other projects.

PR #1465, opened 2025-02-13
fix: resolves Unimplemented type: 3 error for geojson format

rjdjohnston:
We were stuck receiving the unimplemented type 3 error when updating to version 5.x. Moved the isGzipped condition outside the tileJSONformat condition and removed content-encoding headers resolves the issue.

### DIFF
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -117,8 +117,7 @@ export const serve_data = {
       if (isGzipped) {
           data = await gunzipP(data);
           isGzipped = false;
-      }      
-      delete headers['Content-Encoding'];
+      }
       
       if (tileJSONFormat === 'pbf') {
         if (options.dataDecoratorFunc) {
--- a/src/serve_data.js
+++ b/src/serve_data.js
@@ -21,7 +21,7 @@ import { openMbTilesWrapper } from './mbtiles_wrapper.js';
 
 import fs from 'node:fs';
 import { fileURLToPath } from 'url';
-import zlib from 'zlib';
+
 const packageJson = JSON.parse(
   fs.readFileSync(
     path.dirname(fileURLToPath(import.meta.url)) + '/../package.json',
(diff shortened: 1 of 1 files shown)


src/serve_data.js
```diff
@@ -21,6 +21,7 @@ import { openMbTilesWrapper } from './mbtiles_wrapper.js';
 
 import fs from 'node:fs';
 import { fileURLToPath } from 'url';
+import zlib from 'zlib';
```
acalcutt, line 24: Is zlib needed here, we are using gunzipP from

import { gunzipP, gzipP } from './promises.js'; , which links that file itslef

rjdjohnston: You're right. I removed the duplicate import

src/serve_data.js
```diff
@@ -113,12 +114,14 @@ export const serve_data = {
       let headers = fetchTile.headers;
       let isGzipped = data.slice(0, 2).indexOf(Buffer.from([0x1f, 0x8b])) === 0;
 
+      if (isGzipped) {
+          data = await gunzipP(data);
+          isGzipped = false;
+      }      
+      delete headers['Content-Encoding'];
```
acalcutt, line 121: do you think this delete line is needed, won't it just be updated to gzip anyway on line 159 
`headers['Content-Encoding'] = 'gzip';`

rjdjohnston: You are correct. updated.

src/serve_data.js
```diff
@@ -259,8 +261,8 @@ export const serve_data = {
 
         let data = fetchTile.data;
         var param = {
-          long: bbox[0].toFixed(7),
-          lat: bbox[1].toFixed(7),
+          long: bbox[0],
+          lat: bbox[1],
```
acalcutt, line 265: Is this change intentional. I think we purposely limited this to 7, but if it is causing an issue I am fine with reverting it.

acalcutt: any update on this? If this is not needed and can be reverted I will merge moving the ungzip part of this.

acalcutt (approved):

Issue #1488, opened 2025-04-14
tileserver-gl-light is broken. maplibregl missing since version 5.2.0


DVLP:
I don't have full tileserver-gl and I ran a simple example with tileserver-gl-light 5.2.0

Upon opening localhost:8080, getting this

basic-preview/?vector:61 Uncaught ReferenceError: maplibregl is not defined
    at basic-preview/?vector:61:7
﻿
This line in code
```javascript
      maplibregl.setRTLTextPlugin('/mapbox-gl-rtl-text.js' + keyParam);
```

I managed to get it working by downgrading to 5.0.0, then 5.1.3 also worked

Broken version is 5.2.0
Cause: missing maplibregl

This and other errors from the console

![Image](https://github.com/user-attachments/assets/75920cab-0ee8-45e3-bdd7-6a52bc068b43)

acalcutt:
Are you using the docker release or just straight through npm?

In 5.20 all the libraries we changed to be installed as part of prepare https://github.com/maptiler/tileserver-gl/pull/1427 , so those js files are supposed to get copied to the public/resources as part of the npm install

acalcutt:
I tested the docker version of light 5.20 and it seemed to work as explected.

DVLP:
Running directly from npm on Windows

acalcutt:
If you use node v22 does it make a difference? In maplibre-native we only publish binaries for even LTS releases right now.

DVLP:
Let me check, meanwhile I found this.
![Image](https://github.com/user-attachments/assets/7eb3cbc2-ef9e-455a-993b-cc2d4540da76)

DVLP:
Tried with 22 and it's the same as with 23

acalcutt:
I think the issue is the publish.js script is removing the 'prepare' section of the the package.json in https://github.com/maptiler/tileserver-gl/blob/master/publish.js#L40

We need this line now though, because it copies over the all the js files
https://github.com/maptiler/tileserver-gl/blob/master/package.json#L9

When I install light globally I can see that line doesn't exist
![Image](https://github.com/user-attachments/assets/ad315e26-d20d-41a4-bb6c-64b1a21a7f13)

acalcutt:
Can you test this as a workaround?

1.) Open a command prompt to your tileserver-gl folder. when I installed globally it ended up in 
`C:\Users\username\AppData\Roaming\nvm\v22.14.0\node_modules\tileserver-gl-light`

[...]

PR #1491, opened 2025-04-17
Various Updates and Fix RTL Plugin Load

okimiko:
As stated in #1488 the rtl plugin load fails:
- This is caused by the RTL plugin update to version 0.3.0, version 0.2.3 works.
- The plugin load was fixed in [maplibre-gl-js](https://github.com/maplibre/maplibre-gl-js/issues/4850) >= 5

So the option was to downgrade the plugin or upgrade maplibre-gl-js.
This PR goes beyond update just maplibre-gl-js. It updates nearly every compontent (except eslint, sorry :/)


okimiko:
I checked the docker full and light build, vector seems to be OK. But in raster mode, I can see the following messages in case of heavy zooming/movement of the map:
```
GET /styles/bright/512/19/269408/172160.png 200 18424 - 31.289 ms
mlgl: {
  class: 'General',
  severity: 'WARNING',
  text: 'Provided camera options returned 2 tiles, only 14/8419/5379=>19 is taken in Tile mode.'
}
mlgl: {
  class: 'General',
  severity: 'WARNING',
  text: 'Provided camera options returned 2 tiles, only 16/33676/21519=>19 is taken in Tile mode.'
}
mlgl: {
  class: 'General',
  severity: 'WARNING',
  text: 'Provided camera options returned 2 tiles, only 14/8419/5379=>19 is taken in Tile mode.'
}
mlgl: {
  class: 'General',
  severity: 'WARNING',
  text: 'Provided camera options returned 2 tiles, only 16/33676/21519=>19 is taken in Tile mode.'
}
GET /styles/bright/512/19/269408/172161.png 200 12264 - 37.919 ms
```

Dockerfile_build
```diff
@@ -8,6 +8,9 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
       build-essential \
       ca-certificates \
       curl \
+      rsync \
+      jq \
+      unzip \
```
acalcutt, line 13: what are these new packages used for?

okimiko: - `rsync` is used in the publish.js to prepare the light environment
- I used `jq` to check for deprecated packages
  - `jq -r '.packages | to_entries[] | select(.value.deprecated != null) | "\(.key):\n\(.value.deprecated)\n"' package-lock.json ` (https://stackoverflow.com/a/71529396) 
- `unzip` to unpack the test_data

acalcutt: got it. i did not notice this was for the build image so it won't hurt to add these packages.

acalcutt:
I tried this out and it seemed fine to me. the only issue I had is with the elevation control. I am not sure this control is working properly in globe mode, since it give heights outside the globe.

okimiko:
> I tried this out and it seemed fine to me. the only issue I had is with the elevation control. I am not sure this control is working properly in globe mode, since it give heights outside the globe.

OK, may be we need to disable it if Globe is enabled. I will check this later.

acalcutt:
You could also disable the globe button on the terrain preview page for now.

okimiko:
> You could also disable the globe button on the terrain preview page for now.

I checked the output, and I would say the coordinates are "fine" and I would assume it is caused by the rendering of the tiles. Nevertheless I disabled the globe view in the preview until I found out, how to suppress the coordinates or am really sure, that the values are correct.

acalcutt (approved):

PR #1532, opened 2025-04-27
Enable Terrain Preview Elevation API in GlobeView

okimiko:
I added a check to prevent off-globe coordinates (using `map.transform.isPointOnMapSurface(e.point)`) and a marker to show the last clicked position.


acalcutt (approved):

acalcutt:
Just an FYI, I added this issue relating to the markers https://github.com/maptiler/tileserver-gl/issues/1539 , but it is not really something in this code.

PR #1558, opened 2025-05-13
Allow a 'sparse' option per data source

acalcutt:
This PR adds a 'sparse' option to the data source, which can be defined like this in the config file
```
      "Europe_JAXA_SonnyDTM_Italy_France_Switzerland_Merged_Sparse":{
         "mbtiles":"Europe_JAXA_z0-12_SonnyDTM_z0-Z13_Italy_z0-Z14_France_z0-Z15_Switzerland_z0-Z16_Merged_Sparse_cubic.mbtiles",
         "sparse":true
      },
```

When the sparse options is 'true' it changes the error code reported by tileserver-gl when a tile does not exist from a 204 error, which maplibre interprets as no content, to a 410 (gone) error code. The 410 error code allows maplibre to try an fetch tiles from a lower zoom level if the requested tile does not exists instead of loading a blank image like it does with a 204 error.

For the maplibre-native rasters it also does something similar where it sends a empty callback instead of a callback with empty data

This allows tileserver-gl to work with a variable depth sparse tile data source like
https://techidiots.net/downloads/open-base-map-torrents/mbtiles/merged-terrainrgb

The 410 error code was used based on discussion in https://github.com/maplibre/maplibre-gl-js/issues/5692

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,11 +55,6 @@ RUN npm config set maxsockets 1 && \
 
 FROM ubuntu:jammy AS final
 
-ENV \
-    NODE_ENV="production" \
-    CHOKIDAR_USEPOLLING=1 \
-    CHOKIDAR_INTERVAL=500
-
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN export DEBIAN_FRONTEND=noninteractive && \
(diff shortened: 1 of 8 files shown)


acalcutt:
@okimiko  I added some docs for this sparse feature I am trying to added, but i changed the information about terrain data source options in the process. Do you think this looks ok?

https://github.com/maptiler/tileserver-gl/blob/a60cf99275fd04636ec25596b111899da944aa0a/docs/config.rst#data-source-options

Issue #1565, opened 2025-05-16
Strange memory issue raster tile rendering


MichielMortier:
# Memory leak when serving raster tiles

## Description
When using tileserver-gl (v5.3.1) for serving raster tiles, there appears to be a memory leak. The server's memory usage steadily increases when loading raster tiles, unlike with vector tiles where memory usage remains stable.

## Observed Behavior
1. When using vector tiles, memory usage is stable regardless of the number of requests
2. When switching to raster tiles (by changing `vector` to `raster` in the URL), memory usage spikes with each new request
3. Memory increases when viewing different areas/zoom levels that require loading new tiles
4. Memory appears to stabilize after most tiles have been loaded, but is never released
5. When requesting tiles from a new data source, memory usage spikes again

## Evidence
![Memory usage graph](https://github.com/user-attachments/assets/cc5f0003-d002-49b8-b920-9e8cc6997498)

## Environment
- tileserver-gl version: v5.3.1
- Running in Docker: `maptiler/tileserver-gl:v5.3.1`
- Command used: `docker run --rm -it -v "$SCRIPT_DIR/tileserver-gl:/data" -p 8080:8080 maptiler/tileserver-gl:v5.3.1`

## Configuration
**tileserver-gl config.json:**
```json
{
  "options": {
    "paths": {
      "fonts": "fonts",
      "styles": "styles"
    }
  },
  "styles": {
    "tiles-4": {
      "style": "tiles-4/style.json"
    },
    "tiles-5": {
      "style": "tiles-5/style.json"
    }
  },
  "data": {
  }
}
```

**Example style.json:**
```json
{
  "version": 8,
  "name": "Style",
  "sources": {
    "tiles": {
      "type": "vector",
      "tiles": ["https://XXX/{z}/{x}/{y}.pbf"],
      "maxzoom": 14,
      "minzoom": 5
    }
  },
  "sprite": "https://XXX/sprites/osm-bright/icons",
  "glyphs": "https://XXX/fonts/{fontstack}/{range}.pbf",
  "layers": [
    {
      "id": "roads-casing",
      "type": "line",
      "source": "tiles",
      "source-layer": "default",
      "filter": [
        "all", 
        ["has", "testValue"]
      ],
      "layout": {
        "line-cap": "round",
        "line-join": "round",
        "visibility": "visible"
      },
      "paint": {
        "line-color": "#000000",
        "line-width": [
          "interpolate", ["linear"], ["zoom"],
          5, 4,
[...]

PR #1574, opened 2025-05-21
Use jemalloc as memory allocator

MichielMortier:
To fix memory allocations when using raster tiles, I've updated the memory allocator to Jemalloc. 
Discussion can be found [here](https://github.com/maptiler/tileserver-gl/issues/1565)


acalcutt:
Can you fix the dockerfile warning in this. it looks like maybe it wants something quoted.

acalcutt (approved):

PR #1576, opened 2025-05-22
fix the issue of being unable to correctly parse the HTTPS protocol when using an nginx proxy service

dakanggo:
fix the issue of being unable to correctly parse the HTTPS protocol when using an nginx proxy service


dakanggo:
I'll make the fix and commit again

Issue #1669, opened 2025-08-04
Wrong Magic Number Error


thundersoft90:
Hi,

I am a junior map developer and trying to use your docker implementation to open world's mbtiles file downloaded from protomaps' web site. I downloaded config file for your zurich mbtiles file but couldn't use for the another mbtiles file. When I run it via sudo docker run --rm -it -v $(pwd):/data -p 1990:8080 maptiler/tileserver-gl:v5.4.0-pre.1 command, I encountered with the error as follows:

<img width="411" height="77" alt="Image" src="https://github.com/user-attachments/assets/656db8f1-d983-4693-8396-1688113421c2" />

Would you please help me to get rid of this problem ?

Thanks in Advance,

acalcutt:
This usually means an issue with one of you pmtiles files

zstadler:
It would be nice to have the name of the problematic pmtiles file mentioned in the error message...

acalcutt:
I've been meaning to add the pmtiles file name for a while, so thanks for the PR. I agree it is annoying not having that name in the error.

PR #1721, opened 2025-09-15
Add file info to "Wrong magic number" error

zstadler:
Related to #1669


acalcutt:
Looks good to me. I tested with a mbtiles file I renamed to pmtiles and it seems to work and show the name when using either --file or --config options now

image

PR #1860, opened 2025-12-07
Fix get elevation may return data from wrong point - updated

acalcutt:
Update https://github.com/maptiler/tileserver-gl/pull/1759


russellporter (approved):
get elevation endpoint changes lgtm 👍

PR #2158, opened 2026-04-20
fix: clean stale tile-source state on SIGHUP reload

bvitlas:
SIGHUP reload currently stops the HTTP server, but old local tile-source state is not fully cleaned up before the new runtime is started. For MBTiles/PMTiles sources this can leave stale file descriptors alive across reloads, and increase the RAM pressure over time.

In the Docker repro, sending the multiple SIGHUP reloads caused the process to accumulate extra open MBTiles file descriptors using the test data, instead of returning to the baseline. Before the fix, the repro grew from roughly 23 total fds / 4 MBTiles fds to 31+ total fds / 12 MBTiles fds. This suggests old runtime state was surviving multiple SIGHUP reloads.

| State | Total FDs | MBTiles FDs | Deleted FDs |
| --- | ---: | ---: | ---: |
| Before fix, baseline | 23 | 4 | 0 |
| Before fix, after 6 sequential reloads | 31 | 12 | 0 |
| After fix, baseline | 23 | 4 | 0 |
| After fix, after sequential reloads | 23 | 4 | 0 |

---
The screenshot below shows the before-fix repro where sequential reloads increased open MBTiles descriptors.

image

---
This patch fixes the reload lifecycle by:

- Closing old MBTiles-backed sources before replacing the runtime state
- Clearing the PMTiles cache before starting the next runtime generation
- Serializing SIGHUP reload handling so a second reload waits for the current reload to finish

After the fix, the same repro stayed stable at the baseline FD counts after repeated reloads.

## Repro evidence

The repro checked:

```sh
ls /proc/1/fd | wc -l
ls -l /proc/1/fd | grep mbtiles | wc -l
ls -l /proc/1/fd | grep mbtiles | sed 's#.* -> ##' | sort | uniq -c
ls -l /proc/1/fd 2>/dev/null | grep deleted | wc -l
grep -E 'VmRSS|VmSwap|Threads' /proc/1/status
```

### Tested with
- `npm test`
- Docker repro using `tileserver-gl` test data
- Repeated `SIGHUP` reloads while checking `/proc/1/fd` and `/proc/1/status`

### Environment
- `tileserver-gl` Docker image version 5.6.0.
- Test data from the project test dataset
- Reproduced locally with Docker Desktop on Windows
- Similar symptoms were observed in a Linux container as well


acalcutt:
I think overall this seems like a good direction. Can you look at the copilot review and see if any of the fixes mentioned make sense to fix?

bvitlas:
Hi @acalcutt , I addressed these issues reported by copilot.  Check it out. They where mostyl related to error handling so the close/shutdown erros are no longer swallowed. 
 
 As for the reload cleanup: The pending-reload recursion was replaced with a loop, and I added that serveAllStyles watcher is also cleaned up before the next runtime generation starts. 

I also added a small reload test that goes through the SIGHUP handler and checks that cleanup is called for the old runtime. Since `test_data/config.json` does not enable serveAllStyles, I added a separate fixture config  `\test\fixtures\reload-config.json`  for this test instead of changing the shared `test_data/config.json`, because that is what creates the chokidar watcher. Hope that makes sense.

acalcutt (approved):
approved once the merge conflict is fixed

PR #2305, opened 2026-06-30
fix: correctly read allowedHosts from config.json

### LINKED ISSUE #2147: allowHosts not loaded from config
According to #2032 allowed hosts should be loaded in order of presedence config > env > default.

But using the follow config (snippet):
```json
{
  "options": {
    "allowedHosts": "localhost,maps.example.com",
  }
}
```
I still get the warning in the logs. When using the `TILESERVER_GL_ALLOWED_HOSTS` environment variable, it does work correctly though.

artemetra:
This resolves #2147, which I also ran into.
Currently, the issue is that in main.js in `startServer` we attempt to read `allowedHosts` from the `config` variable and not `configPath`.  
By default, if we are not starting with an input file, we run `startServer(opts.config, null);`, where `null` becomes the `config` variable, hence `allowedHosts` is completely ignored.
The solution is to move this logic to where the config is actually read, which is in server.js. 

Let me know what you think and whether this needs more attention.


phknot:
This seems like a proper fix for https://github.com/maptiler/tileserver-gl/issues/2147 can we get a maintainer to merge this?

PR #2344, opened 2026-08-27
fix: align 512px static map overlays

miakh:
## Summary

- select the MapLibre Native zoom from the explicit `tile`/`static` render mode, keeping the 512px exception tile-only
- isolate that policy in a dependency-free helper with coverage for 256/512 tiles and 400/512 static maps
- cover 512px static overlays at both @1x and @2x through the existing endpoint test helper

Fixes #2326

## Testing

- PASS: dependency-free Node assertions for the complete render-mode/width matrix and unsupported modes
- PASS: `node --check src/render_mode.js`, `node --check test/render_mode.js`, and `node --check src/serve_rendered.js`
- PASS: `git diff --check`
- NOT RUN locally: repository Mocha/lint gate; `npm ci --prefer-offline --no-audit --omit=optional` cannot install `@maplibre/maplibre-gl-native` under the host's unsupported Node 26 ABI (the project supports Node 20/22/24 and CI selects from `package.json`)

PR #2355, opened 2026-08-31
fix: align low-zoom static map overlays

### LINKED ISSUE #1021: Map Markers are placed incorrectly at low zoom levels
## **Issue**
Low zoom levels using markers are broken (ie the marker doesn't appear in the correct location), for latitude 

**Working Zoom Level (marker on Zurich)**
```
http://localhost:8080/styles/test-style/static/8.5432,47.3669,10/1000x1000.png?marker=8.5432,47.3669|https://cdn4.iconfinder.com/data/icons/small-n-flat/24/map-marker-512.png|scale:0.1|offset:0,30
```
image

**Zoom Level Broken**
```
http://localhost:8080/styles/test-style/static/8.5432,47.3669,1/1000x1000.png?marker=8.5432,47.3669|https://cdn4.iconfinder.com/data/icons/small-n-flat/24/map-marker-512.png|scale:0.1|offset:0,30
```
image

Config used looks something like this
```json
{
  "options": {
    "paths": {
      "styles": "styles",
      "fonts": "fonts"
    },
    "allowRemoteMarkerIcons": true,
    "allowInlineMarkerImages": true
  },
  "styles": {
    "test-style": {
      "style": "osm-bright/style.json",
      "tilejson": {
        "type": "overlay",
        "bounds": [8.529446, 47.364758, 8.55232, 47.380539]
      }
    }
  },
  "data": {
    "openmaptiles": {
      "mbtiles": "zurich_switzerland.mbtiles"
    }
  }
}

```

### **Thoughts on why this is happening**
I did some investigating. Below is the code that I think needs to be fixed.
https://github.com/maptiler/tileserver-gl/blob/2a441a00391e875d9966939e8db2579df27ea658/src/serve_rendered.js#L338-L380

Basically what I believe is happening is that if the height of the requested image is too large, the map will hit it's max zoom and stop zooming for lower zoom levels, ie notice that the map at zoom level `1` is the exact same as the map at zoom level `1.8` if you request an `1000x1000` image. Because of this, I believe the mercator projection is still expecting the map to be furthered zoomed out, while the zoom of the image is the same and thus causing a mismatch. Not sure what the fix is though.

Curl 1
```
http://localhost:8080/styles/test-style/static/8.5432,47.3669,1/1000x1000.png?marker=8.5432,47.3669|https://cdn4.iconfinder.com/data/icons/small-n-flat/24/map-marker-512.png|scale:0.1|offset:0,30
```

Curl 1.8
```
http://localhost:8080/styles/test-style/static/8.5432,47.3669,1.8/1000x1000.png?marker=8.5432,47.3669|https://cdn4.iconfinder.com/data/icons/small-n-flat/24/map-marker-512.png|scale:0.1|offset:0,30
```
&nbsp;
&nbsp;
&nbsp;

### **About**

Macos: 13.5.1
Docker: 4.24.0 
TileserverGL: 4.5.2

miakh:
## Summary

- align center-based static-map overlays with MapLibre Native's effective minimum viewport zoom
- keep horizontal wrapping and the dedicated 256px zoom-zero render path unchanged
- add unit, request-level, and visual regression coverage for the low-zoom marker case

Fixes #1021.

## Verification

- `npx prettier --check test/static_images.js`
- `npx mocha test/setup.js test/render_mode.js test/static.js test/static_images.js --grep 'Static overlay zoom|keeps overlays aligned|static-marker-low-zoom-large-viewport' --timeout 20000 --exit` (6 passing)
- `git diff --check upstream/master...HEAD`


acalcutt:
Can you make a static image test similar to the last PR? I think there is already a marker example in there, so it would just need the low zoom version of it

miakh:
Yes — I’ll add a visual regression case using the existing marker fixture at low zoom with a larger viewport, similar to #2344. I’ll include the expected image in the next update.

miakh:
Added the low-zoom visual regression in be40621 using the existing marker fixture at zoom 1 with a 1000×1000 viewport. The expected image is committed with the test:

Expected low-zoom marker rendering

The branch is also updated onto the current `master`; the targeted regression checks pass locally (6 passing).

